PR #34, opened 2026-06-10
Similar notes in the note context sidebar

maccman:
## What

Ordinary notes now get a right-hand context sidebar — the old app's note context sidebar — showing the note's inbound links ("Linked from") and its semantic neighbors ("Similar notes"), powered by the Plan 09 local embeddings.

- **`NoteContextSidebar`** mounts in the AppShell's right region on `note` routes, alongside the existing daily sidebar on daily routes. Route → sidebar mapping lives in `context-sidebar/sidebar-route.ts` (`contextSidebarFor`).
- **No double rendering:** the note pane's inline backlinks/similar panels hide at the `lg` breakpoint where the sidebar appears (`contextInSidebar` prop), so the context shows exactly once at every window size. The daily stream keeps its inline panels because its sidebar describes only the target day, not every visible one.
- **V1 quality threshold:** `relatedNotes()` now drops candidates past the old app's 0.7 cosine-distance cutoff instead of padding the list with KNN's nearest-whatever (worst in small graphs).
- **Cosine metric stored natively:** migration `0003_cosine_vectors.sql` rebuilds `embedding_vectors` as `vec0(embedding float[384] distance_metric=cosine)`, copying the stored vectors as-is (no re-embedding; KNN ordering is unchanged since the MiniLM vectors are unit-normalized). The threshold applies directly to vec0 distances, and retrieval scores become true cosine similarity. Two new Rust db tests pin the metric (a same-direction, half-magnitude probe must sit at distance 0 — L2 would say 0.5) and prove vectors staged on the 0002 schema survive the rebuild.
- **Shared hooks + sections:** `useBacklinks`/`useSimilarNotes` own the index query keys (the contract with the invalidation hook), consumed by the inline panels and the new path-based `BacklinksSection`/`SimilarNotesSection`, which both sidebars share.
- **Naming:** every user-facing "Related" heading is renamed to the old app's "Similar notes".

## Why

Plan 09 shipped the retrieval API and a flat list under the editor, but regular notes had no context sidebar at all — and without a similarity cutoff, sparse graphs surfaced unrelated notes as "similar". Storing cosine distances natively keeps the cutoff's meaning in the schema instead of coupling it to fastembed's normalization behavior in TypeScript.

## Reviewer notes

- The threshold changes results on all three surfaces (note sidebar, daily sidebar, inline list) — intentionally: fewer-but-right beats always-six.
- Migration 0003 is append-only per the migrations module's contract; fresh installs run it against an empty table.
- Collapse state keys changed prefix, so any collapsed sections reset once (sessionStorage only).
- Verified: `pnpm typecheck`, `pnpm lint`, core + desktop vitest suites (52 tests), `cargo test db::` (23 tests), `cargo fmt --check`, `cargo clippy` — all green.


maccman:
Squash-merged into master via the consolidation PR #39.

PR #48, opened 2026-06-10
Filter semantic search results by the cosine noise cutoff

maccman:
## What changed

Semantic search returned results for any query — even gibberish like `xxxxxxxxxxxx` that appears nowhere in the graph. The sqlite-vec KNN query always fills its candidate list with the nearest `k=24` chunks however distant they are; the lexical (FTS) leg correctly found nothing, but `semanticHits()` served the unfiltered neighbors, so hybrid search showed pure noise.

`relatedNotes()` already guarded against exactly this with a 0.7 cosine-distance cutoff (`MAX_RELATED_COSINE_DISTANCE`); search never applied one.

- Extracted the near-identical dedupe loops from `semanticHits()` and `relatedNotes()` into a shared, pure `bestChunkPerNote()` helper that drops rows past the cutoff, collapses to the best chunk per note, and (for related notes) excludes the seed note.
- Renamed the constant to `MAX_COSINE_DISTANCE` since it now governs both paths.
- Added unit tests for the cutoff, per-note dedup, seed exclusion, limit, snippet trimming, and the similarity/private-flag conversions.

## Why 0.7 works for search too

The 0.7 value was tuned for note→note similarity; I verified it also separates query→chunk matching cleanly by running the same model (all-MiniLM-L6-v2) against realistic note chunks:

| Query type | Best-match cosine distance |
|---|---|
| Relevant ("vacation ideas", "taxes", "vector search sqlite") | 0.30 – 0.65 |
| Fuzzy-related ("baking" vs a sourdough note) | 0.49 – 0.63 |
| Gibberish (`xxxxxxxxxxxx`, `qqqqqqqq`, …) | 0.72+ |
| Real-but-unrelated ("quantum chromodynamics") | 0.83+ |

Margin on both sides of 0.7, so no behavior change for legitimate queries. These numbers are recorded in the constant's doc comment.

## Notes for reviewers

- AI callers using `retrieve()` (semantic or hybrid mode) also stop receiving junk context for off-topic questions — same code path.
- Hybrid behavior for gibberish is now: lexical finds nothing, semantic filters everything out → empty results, as expected.
- Verified with `pnpm check` and `pnpm test --run src/embeddings/retrieve.test.ts` (9 tests pass).

---

> [!NOTE]
> **Medium Risk**
> Changes shared retrieval behavior for semantic/hybrid search and AI context; legitimate edge-case queries near the 0.7 distance boundary could lose hits, but the threshold was already validated for related notes.
> 
> **Overview**
> Semantic search and hybrid retrieval no longer return KNN neighbors when every chunk is past the **0.7 cosine-distance** noise threshold—gibberish or off-topic queries can return empty results instead of the nearest unrelated chunks.
> 
> The dedupe/cutoff logic shared by **`semanticHits`** and **`relatedNotes`** is moved into a pure **`bestChunkPerNote()`** helper (drop distant rows, keep the best chunk per note, optional seed exclusion, similarity scoring). The constant is renamed to **`MAX_COSINE_DISTANCE`** and documented for query→chunk behavior. Unit tests cover the cutoff, deduping, limits, seed exclusion, and snippet/private handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b4e99f286514c3ee3a7109d822e9454628b0d92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/embeddings/retrieve.test.ts
+++ b/packages/core/src/embeddings/retrieve.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { fuseRanked, type RetrievalHit } from './retrieve'
+import { bestChunkPerNote, fuseRanked, type ChunkHitRow, type RetrievalHit } from './retrieve'
 
 function hit(path: string, overrides?: Partial<RetrievalHit>): RetrievalHit {
   return {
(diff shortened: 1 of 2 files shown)


maccman:
@coderabbitai review

PR #85, opened 2026-06-11
Use shadcn Dialog, Button, and Input where we were hand-rolling them

maccman:
## Summary

- **`add-ai-model-dialog`** — biggest change: replaced ~40 lines of hand-rolled modal (fixed overlay, manual focus trap, Tab-cycle keyboard handler, focus-restore `useEffect`, manual `role="dialog"` / `aria-modal`) with `Dialog` + `DialogContent` from `@/components/ui/dialog`. Radix handles all of that for free. Also replaced the Cancel/Submit `<button>` elements with `Button` and the API key `<input>` with `Input`. Native `<select>` elements are kept intentionally so the existing `fireEvent.change`/`getByLabelText` tests keep working without refactor.
- **`ai-models-section`** — "Add model" raw `<button>` → `Button variant="ghost" size="sm"`
- **`ai-model-row`** — "Make default" and trash icon raw `<button>` elements → `Button variant="ghost"`
- **`graph-chooser`** — primary "Open graph…" CTA and "Forget" raw `<button>` elements → `Button`

## What was audited but left alone

| Component | Reason kept custom |
|---|---|
| `inline-alert.tsx` | 37-line component using DS tokens directly; shadcn `Alert` adds unused icon slots |
| `model-download-progress.tsx` | Handles indeterminate (pulse) + determinate in one pass; shadcn `Progress` doesn't cover the indeterminate shimmer case natively |
| `filter-tab.tsx` | Filter toggles, not traditional tabs; `Tabs` semantics don't fit |
| `settings/option-card.tsx` | Custom radio card with design tokens; `RadioGroup` would lose the card-wraps-anything abstraction |

## Test plan

- [ ] All 11 `ai-models-section` tests pass (Dialog role, label associations, Tab trap, add/remove/default flows)
- [ ] Both `graph-chooser` tests pass
- [ ] Manually open Settings → AI Models, add a model through the dialog; verify focus, Escape, and outside-click all close it correctly
- [ ] Verify "Open graph" and "Forget" buttons in the first-run screen look and function correctly

---

> [!NOTE]
> **Low Risk**
> UI-only refactor with no changes to API key validation, keychain storage, or model settings logic; main regression surface is dialog focus/keyboard behavior, which retains an explicit focus-restore workaround.
> 
> **Overview**
> Replaces hand-rolled desktop UI primitives with shared **shadcn** components for consistent styling and accessibility, without changing AI model persistence or key verification behavior.
> 
> The largest change is **`add-ai-model-dialog`**: the custom overlay, `role="dialog"` markup, Escape/outside-click handling, and manual Tab focus trap are removed in favor of **`Dialog` / `DialogContent`**. Cancel/submit actions and the API key field now use **`Button`** and **`Input`**; native **`<select>`** controls stay as-is to keep existing test interactions. A mount/unmount **`useEffect`** still restores focus to the opener because the dialog is conditionally mounted and direct `onClose()` can bypass Radix auto-focus.
> 
> **`graph-chooser`**, **`ai-models-section`**, and **`ai-model-row`** swap raw `<button>` elements for **`Button`** variants on the primary open action, “Forget”, “Add model”, “Make default”, and remove controls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cbd18ae8efbaa82e9d2e093aadb3cc1d4462542. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/desktop/src/components/graph-chooser.tsx
+++ b/apps/desktop/src/components/graph-chooser.tsx
@@ -1,5 +1,6 @@
 import { type ReactElement } from 'react'
 import { Folder, FolderPlus } from 'lucide-react'
+import { Button } from '@/components/ui/button'
 import { useGraph } from '@/providers/graph-provider'
 
 /**
@@ -19,14 +20,14 @@ export function GraphChooser(): ReactElement {
           </p>
         </div>
 
-        <button
+        <Button
           type="button"
+          className="w-full"
           onClick={() => void pickAndOpen()}
-          className="flex w-full items-center justify-center gap-2 rounded-md bg-accent px-3 py-2 text-sm font-medium text-text-on-brand shadow-sm transition-colors duration-100 hover:bg-accent-hover"
         >
-          <FolderPlus aria-hidden strokeWidth={1.75} className="size-4" />
+          <FolderPlus aria-hidden strokeWidth={1.75} />
           Open graph…
-        </button>
+        </Button>
 
         {error ? (
           <p role="alert" className="text-center text-sm text-destructive">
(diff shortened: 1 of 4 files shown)


maccman:
Addressed the focus-restore bug from comment 3397278549: added back the useEffect that captures document.activeElement at mount and restores it in cleanup. Radix's onCloseAutoFocus only fires through its own close path — when Cancel or successful submit calls onClose() directly the parent unmounts the whole component tree, bypassing Radix's Presence flow. The React cleanup fires on every unmount regardless.

PR #405, opened 2026-06-26
Fix Task View sync after note-side task reopen

maccman:
## Context

Task View deliberately keeps a session-only middle state: after you complete a task from Task View, the row stays visible and struck through until you archive it. That mirrors the V1 workflow and gives users a quick undo path.

The support report exposed the missing reconciliation path. If the same checkbox is reopened from the source note, such as a daily note, the index eventually reports that task as open again, but Task View was still letting the session-completed `[x]` copy shadow the real `[ ]` row. The row kept offering **Reopen**, and clicking it attempted to find the old completed line in markdown that had already changed, producing the “task line no longer in note” toast.

## What Changed

- Adds `forgetRecentlyCompletedNowOpen`, which reconciles the session-completed set against the latest open-task index rows.
- Removes a struck session copy only when the indexed open row has the same task identity and a strictly newer `updatedAt`.
- Runs that reconciliation in `TasksScreen` whenever the open-task query updates, so note-side reopen wins once the index has actually caught up.
- Preserves the intended Task View middle state when an open-task refetch is stale or same-version noise.
- Adds regression coverage for both the store rule and the full screen flow.

## User-Facing Behavior

After this PR:

- Completing a task in Task View still leaves it struck through until Archive.
- Reopening the same task in its source note now makes Task View show it as open again.
- The next Task View action uses the current `[ ]` markdown row instead of a stale `[x]` row, avoiding the support-channel failure mode.

## Verification

Passed in the linked worktree:

- `./node_modules/.bin/vitest run src/components/tasks/tasks-screen.test.tsx` (57 tests)
- `./node_modules/.bin/vitest run src/lib/tasks/recently-completed.test.ts` (5 tests)
- `./node_modules/.bin/tsc --noEmit` from `apps/desktop`
- `./node_modules/.bin/oxlint apps packages` (exit 0; existing warnings only)
- `./node_modules/.bin/eslint .` from `apps/desktop` (exit 0; existing warnings only)
- `./node_modules/.bin/vite build` from `apps/desktop` (passes; existing chunk-size warnings only)
- `git diff --check`

`pnpm check` could not start in this linked worktree because pnpm attempted to remove `node_modules` without a TTY. To avoid mutating the shared dependency install, I ran the underlying typecheck, lint, and build checks directly.

## Risk and Rollout

This is a low-risk UI/session-state change:

- No markdown format changes.
- No index migrations.
- No persistence changes.
- No external service calls.

The reconciliation depends on `updatedAt` advancing when the note-side reopen is indexed. If a platform failed to advance that timestamp, Task View could keep the struck session row until a later refresh or archive, but the existing markdown write guard would still prevent data loss. Rollback is a normal revert of this branch.

### DIFF
--- a/apps/desktop/src/components/tasks/tasks-screen.test.tsx
+++ b/apps/desktop/src/components/tasks/tasks-screen.test.tsx
@@ -204,13 +204,16 @@ function RouteProbe(): ReactNode {
 
 function renderScreen() {
   const client = new QueryClient({ defaultOptions: { queries: { retry: false } } })
-  return render(
-    <QueryClientProvider client={client}>
-      <RouterProvider>
-        <TasksScreen />
-        <RouteProbe />
-      </RouterProvider>
-    </QueryClientProvider>,
+  return Object.assign(
+    render(
+      <QueryClientProvider client={client}>
+        <RouterProvider>
+          <TasksScreen />
+          <RouteProbe />
+        </RouterProvider>
+      </QueryClientProvider>,
+    ),
+    { client },
   )
 }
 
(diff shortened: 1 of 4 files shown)


maccman:
I'm gonna wait until Fable is out and we can redo this one.

PR #425, opened 2026-06-28
fix(desktop): keep daily stream anchored during note load

ocavue:
Just for doc reason. do not merge.


ocavue:
Close in favor of https://github.com/team-reflect/reflect-open/pull/426

PR #439, opened 2026-06-30
Support opening graph folders from Dock drops

vojto:
## Summary
- Advertise macOS folder support in the app bundle so the Dock icon accepts folder drops.
- Queue native open requests only when exactly one dropped item is a directory; ignore files and multi-item drops.
- Drain queued requests through GraphProvider so cold launches and running apps use the normal graph-open lifecycle.

## Tests
- `pnpm --filter @reflect/desktop test src/providers/graph-provider.test.tsx`
- `pnpm check`
- `pnpm --filter @reflect/desktop sidecar`
- `cargo test -p reflect-open graph_open_request`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added an IPC-driven “graph open request” API and event to handle native folder opens.
  * Introduced a Dock/Finder hook to drain queued folder open requests and open the corresponding root in the app.
  * Updated macOS to declare a dedicated document type for “Reflect Graph Folder.”
* **Bug Fixes**
  * Improved launch and in-app handling so queued native folder opens are processed before falling back to the most recent graph.
* **Tests**
  * Added coverage to verify queued native folder requests are prioritized and handled correctly during startup and while running.

### DIFF
--- a/apps/desktop/src-tauri/Info.plist
+++ b/apps/desktop/src-tauri/Info.plist
@@ -5,6 +5,21 @@
      release builds. -->
 <plist version="1.0">
 <dict>
+	<key>CFBundleDocumentTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeName</key>
+			<string>Reflect Graph Folder</string>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>LSHandlerRank</key>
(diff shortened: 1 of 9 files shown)


maccman:
Very much appreciate the contribution here, but this pr makes less sense now with iCloud.

PR #467, opened 2026-07-02
feat: enforce the notes kind/daily_date invariant at the schema level

maccman:
## Problem

Migration 0014 added `notes.kind` (`'daily' | 'note' | 'template'`), which redundantly encodes "is this a daily note" alongside the older `daily_date` column — both are derived from the path in `buildIndexedNote`, so they agree only by construction. Nothing made SQLite reject a drifted writer: a row with `kind = 'note'` and a `daily_date`, or `kind = 'daily'` with none, would be silently accepted and then surface inconsistently, because queries filter on one column or the other interchangeably (All Notes uses `kind`, the daily-date indexes and `note_keys` use `daily_date`).

0014 already shipped in the tagged v0.4.0-beta.12, so the constraint lands as an appended migration 0015 rather than a rewrite of 0014.

## Changes

1. **Migration `0015_note_kind_invariant.sql`** — adds `CHECK ((kind = 'daily') = (daily_date IS NOT NULL))` to `notes`. SQLite cannot `ADD` a table-level CHECK, so the table is dropped and recreated with the constraint, following the same wipe-for-reindex pattern as 0004/0006/0014: the projection rows are deleted children-first, then `notes` is dropped and recreated under its own name, and the seven `notes` indexes (0001/0007/0010/0013) are recreated. Deleting the children first is what makes the `DROP TABLE` safe under the app's runtime `PRAGMA foreign_keys = ON` — the drop's implicit `DELETE FROM` cascades over already-empty tables. The child tables' `REFERENCES` clauses and the `note_keys`/`backlinks` views bind to `notes` by name, so they resolve against the recreated table unchanged. `index_meta`, the content-hash-keyed embedding tables, and the durable `chat_*` tables are untouched; the next open re-indexes everything, exactly like the previous wipe migrations.
2. **`crates/index-schema/src/lib.rs`** — appends the 0015 migration and bumps `LATEST_SCHEMA_VERSION` to 15. No behavioral changes to the migration runner.
3. **Fixture fixes** — two raw-insert fixtures wrote rows the CHECK now rejects: the CLI's `build_index` (`apps/cli/tests/cli.rs`) inserted daily rows while leaving `kind` at its `'note'` default and now derives `kind` from the daily date, and the desktop open-tasks test staged a `notes/a.md` row with a `daily_date` and now stages a real daily note.

No TS changes: `PROJECTION_VERSION` only moves when row derivation changes, and it doesn't — this constrains what SQLite accepts, not what the indexer writes.

## Tests

- `note_kind_daily_date_invariant_is_enforced` (`db/tests.rs`) — the CHECK rejects `kind='note'` with a `daily_date`, `kind='daily'` without one, and `kind='template'` with one; the three shapes `buildIndexedNote` actually emits all insert cleanly.
- `kind_invariant_migration_wipes_the_projection_for_reindex` (`db/tests.rs`) — stages parent + child + FTS + `index_meta` rows at schema v14 with foreign keys ON, migrates to latest, and asserts: the projection is wiped (so the DROP ran without FK errors), `index_meta` survives, all seven indexes exist on the recreated table, and `apply_note`/`backlinks`/`ON DELETE CASCADE` all resolve against it.
- Existing version pins updated (14 → 15) in `migrations_are_valid_and_idempotent` and `open_index_at_creates_migrates_and_reopens`.

## Verification

```
cargo test -p reflect-index-schema   # 2 passed
cargo test -p reflect-open           # 139 passed (sidecars staged first)
cargo test -p reflect-cli            # 54 passed
cargo clippy -p reflect-index-schema -p reflect-open -p reflect-cli --all-targets   # clean
cargo fmt --check                    # clean
```

## Risk / Rollout

- The migration wipes the note projection, so every graph pays a full re-index on next open — the same cost 0014 charged in this release train, and the standard recovery path (0004/0006/0009/0014 all do this). `chat_*` history and the content-hash-keyed embeddings survive, so nothing durable is lost and unchanged notes are not re-embedded.
- Unlike a copy-based rebuild, the migration cannot fail on a pre-existing inconsistent row — there is no path where a drifted row bricks graph-open.

### DIFF
--- a/apps/cli/tests/cli.rs
+++ b/apps/cli/tests/cli.rs
@@ -40,7 +40,11 @@ impl Fixture {
             let content = fs::read_to_string(self.root().join(&note.rel_path)).unwrap();
             let meta = parse_note_meta(&note.rel_path, &content);
             let daily_date = reflect_cli::paths::date_from_daily_path(&note.rel_path);
-            let kind = if daily_date.is_some() { "daily" } else { "note" };
+            let kind = if daily_date.is_some() {
+                "daily"
+            } else {
+                "note"
+            };
             conn.execute(
                 "INSERT INTO notes(path, id, title, title_key, kind, daily_date, is_private,
                                    is_pinned, pinned_order, file_hash, mtime, updated_at, preview)
--- a/apps/cli/tests/cli.rs
+++ b/apps/cli/tests/cli.rs
@@ -40,10 +40,11 @@ impl Fixture {
             let content = fs::read_to_string(self.root().join(&note.rel_path)).unwrap();
             let meta = parse_note_meta(&note.rel_path, &content);
             let daily_date = reflect_cli::paths::date_from_daily_path(&note.rel_path);
+            let kind = if daily_date.is_some() { "daily" } else { "note" };
             conn.execute(
-                "INSERT INTO notes(path, id, title, title_key, daily_date, is_private, is_pinned,
-                                   pinned_order, file_hash, mtime, updated_at, preview)
-                 VALUES(?1, ?8, ?2, ?3, ?4, ?5, 0, NULL, ?6, ?7, ?7, '')",
+                "INSERT INTO notes(path, id, title, title_key, kind, daily_date, is_private,
+                                   is_pinned, pinned_order, file_hash, mtime, updated_at, preview)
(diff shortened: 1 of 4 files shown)


maccman:
@coderabbitai review

PR #612, opened 2026-07-06
fix(ios): preserve the ShareExtension App Group entitlement in CI TestFlight builds

maccman:
## Problem

Sharing a page to Reflect from Safari or Chrome on a TestFlight install fails every time: the share sheet opens, then shows "Couldn't save". The same build works from a locally-built IPA, and debug simulator builds work too — only CI-built TestFlight IPAs are broken.

Root cause: Tauri's App Store Connect API-key flow (what CI uses) archives with `CODE_SIGNING_ALLOWED=NO`, then dummy-signs **only the app binary** so its entitlements survive export (`tauri-cli` `mobile/ios/build.rs`). `xcodebuild -exportArchive` preserves only entitlements it finds in an existing signature, so the unsigned `ShareExtension.appex` is re-signed with **no App Group entitlement**. At runtime `containerURL(forSecurityApplicationGroupIdentifier:)` returns nil, the spool throws `containerUnavailable`, and every share fails. Local builds are unaffected because without the API-key env vars Xcode signs the appex (entitlements included) during archive.

Verified end to end:
- Simulator: an entitlement-stripped release appex reproduces "Couldn't save" exactly; the same appex ad-hoc signed *with* entitlements saves fine.
- Local replay of CI: exporting an unsigned device archive (`tauri ios build --archive-only --no-sign --ci`) drops the App Group from the appex; pre-signing the appex preserves it through export with the full correct entitlement set.
- Real CI: run [28820291156](https://github.com/team-reflect/reflect-open/actions/runs/28820291156) (this branch, before the fix commit) fails the new assertion with `ShareExtension.appex is signed without the group.app.reflect App Group entitlement`.

## Before → After

| | Before | After |
|---|---|---|
| Share to Reflect on a TestFlight build | "Couldn't save", capture lost | Saves; envelope reaches the App Group inbox |
| CI builds a broken appex | Uploads to TestFlight silently | `release-ios` fails loudly before upload |
| TestFlight IPA bytes | Not retrievable after upload | Kept as a workflow artifact |
| `xcodegen`/`tauri ios init` regen | Wipes both `.entitlements` files + drops Info.plist usage strings | Idempotent — contents are spec-driven |

## Changes

1. **`ios.project.yml` / `gen/apple/project.yml` / `project.pbxproj`** — a post-build phase on the ShareExtension target ("Preserve entitlements on skip-codesign builds") ad-hoc signs the appex with its entitlements when `CODE_SIGNING_ALLOWED=NO`. This mirrors what Tauri already does for the app binary; export then re-signs both with real certs and the App Group intact. Properly signed builds never enter the branch.
2. **Spec-driven entitlements** — entitlement contents now live under `entitlements.properties` in the project spec (and the mic/calendar usage descriptions under `info.properties`), so regenerating the project rewrites `reflect-open_iOS.entitlements` / `ShareExtension.entitlements` / `Info.plist` correctly instead of clobbering them — retiring the long-standing "restore entitlements by hand after `tauri ios init`" foot-gun.
3. **`release-ios.mjs`** — `assertIpaAppexEntitlements` reads the signed entitlements of every embedded appex (`codesign -d --entitlements`) and refuses to accept or upload an IPA missing `group.app.reflect`; runs both post-build and pre-upload. `findIpaAppexPaths` is exported and unit-tested.
4. **`testflight.yml`** — uploads the exported IPA as a workflow artifact (`if: always()`), since TestFlight builds can't be downloaded back for inspection.
5. **`docs/porting/reflect-mobile/share-extension.md`** — records the CI-signing pitfall and the new invariants.

## Tests

- `apps/desktop/scripts/release-ios.test.mjs` — appex path discovery in an IPA listing (single extension found once; IPAs without extensions return empty).
- Manual matrix (simulator, iPhone Air / iOS 26.5): debug appex ✅, unsigned release appex ❌ (`containerUnavailable`, reproduces the field failure), ad-hoc+entitlements release appex ✅ (envelope written to the App Group inbox).
- Export-preservation A/B on a real unsigned device archive: patched appex keeps the App Group through `-exportArchive`; unpatched control loses it (byte-for-byte the broken TestFlight entitlement set).

## Verification

- `pnpm check` — clean.
- `pnpm test --run scripts/release-ios.test.mjs` — 15 passed.
- `xcodebuild -target ShareExtension -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO` on the regenerated project — build succeeds and the phase leaves the appex ad-hoc signed with `group.app.reflect`.
- CI on this branch: [28820291156](https://github.com/team-reflect/reflect-open/actions/runs/28820291156) fails the assertion pre-fix (bug proven on the real pipeline); [28821066001](https://github.com/team-reflect/reflect-open/actions/runs/28821066001) is the post-fix TestFlight run — expected to pass the assertion and upload a working build (in progress at PR time).

## Risk / Rollout

- The signing phase is a no-op for every normally-signed build (local dev, `tauri ios dev`, local TestFlight builds); it only touches skip-codesign builds.
- If Apple's export ever stops preserving ad-hoc-carried entitlements, the new assertion fails the build loudly instead of shipping another broken extension.
- A device pass on the next TestFlight build is still owed (share from Safari + Chrome on iPhone) — the fix is proven at the entitlement level; the on-phone confirmation closes the loop.
- Upstream: this is a Tauri bug (`tauri-cli` dummy-signs only the app binary, never embedded appexes) — worth filing so the workaround can eventually be dropped.

---

> [!NOTE]
> **Medium Risk**
> Touches iOS code signing, entitlements, and the TestFlight release gate; mistakes could break share capture or block uploads, but changes are scoped with pre-upload assertions and a no-op for normally signed builds.
> 
> **Overview**
> Fixes **TestFlight shares failing with "Couldn't save"** by keeping `group.app.reflect` on `ShareExtension.appex` through Tauri's CI export path, where signing is skipped and export only carried entitlements already on a signature (the main app was dummy-signed; the extension was not).
> 
> A **ShareExtension post-build phase** ad-hoc signs the appex with `ShareExtension.entitlements` when `CODE_SIGNING_ALLOWED=NO`; normal local/Xcode-signed builds skip that branch. **`ios.project.yml` / `gen/apple/project.yml`** now declare entitlements and mic/calendar usage strings under `entitlements.properties` and `info.properties` so xcodegen/`tauri ios init` regens no longer wipe `.entitlements` or drop those plist keys.
> 
> **`release-ios.mjs`** adds `findIpaAppexPaths` and `assertIpaAppexEntitlements` (via `codesign -d --entitlements`) inside `assertIpaAppStoreMetadata`, blocking build/upload/validate if the appex lacks the App Group. The **TestFlight workflow** uploads the exported IPA as an artifact (`if: always()`). Porting docs record the CI-signing pitfall and spec-driven entitlements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff9300db4bea128f1c4efa0bf9e43ffc852c9744. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -91,3 +91,13 @@ jobs:
         run: |
           pnpm release:ios preflight --build-number="$BUILD_NUMBER"
           pnpm release:ios testflight --build-number="$BUILD_NUMBER" --export-method="$EXPORT_METHOD" $WAIT_FLAG
+
+      # The exported IPA, kept for post-hoc signing/entitlements inspection —
+      # TestFlight builds cannot be downloaded back from App Store Connect.
+      - name: Upload IPA artifact
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: reflect-ios-ipa
+          path: apps/desktop/src-tauri/gen/apple/build/arm64/*.ipa
+          if-no-files-found: ignore
--- a/apps/desktop/scripts/release-ios.mjs
+++ b/apps/desktop/scripts/release-ios.mjs
@@ -24,6 +24,7 @@ const IOS_BUNDLE_IDENTIFIER = 'app.reflect.ios'
 const OLD_CAPACITOR_BUNDLE_IDENTIFIER = 'app.reflect.ReflectMobile'
 const NON_EXEMPT_ENCRYPTION_KEY = 'ITSAppUsesNonExemptEncryption'
 const KEYCHAIN_SERVICE = 'reflect-notary'
+const SHARE_EXTENSION_APP_GROUP = 'group.app.reflect'
 
 const here = dirname(fileURLToPath(import.meta.url))
 const appDir = join(here, '..')
@@ -126,6 +127,16 @@ export function findIpaInfoPlistPath(listing) {
 /** Return true when an Info.plist raw value represents boolean false. */
 export function isFalsePlistValue(value) {
   return ['false', 'no', '0'].includes(value.trim().toLowerCase())
+}
+
+/** Find every app-extension bundle path inside an IPA's unzip listing. */
+export function findIpaAppexPaths(listing) {
+  const paths = new Set()
+  for (const line of listing.split('\n')) {
+    const match = line.trim().match(/^(Payload\/[^/]+\.app\/PlugIns\/[^/]+\.appex)\//)
+    if (match) paths.add(match[1])
+  }
+  return [...paths].sort()
 }
 
 function ensureMacos() {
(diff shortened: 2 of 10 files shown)


maccman:
Upstream issue filed: tauri-apps/tauri#15663 — `tauri-cli`'s skip-codesign flow dummy-signs only the app binary, so embedded appexes lose their entitlements at export. Once fixed upstream, the "Preserve entitlements on skip-codesign builds" phase can be dropped (the `release-ios.mjs` assertion should stay as the guard).

PR #627, opened 2026-07-07
feat(mobile): double-tap All to search, return key dismisses the keyboard

maccman:
## Problem

On iOS, getting into search takes two precise taps: the All tab, then the small search field at the top of the screen. The Daily tab already has a faster capture gesture — double-tapping it focuses today's editor — but the All tab had no equivalent, so the most common reason to visit it (searching) had no fast path.

Once typing, there was also no good way to put the keyboard away: the keyboard's blue return key did nothing (the All and Tasks lists both filter live, so there was nothing to submit), and the half-covered result list stayed half-covered.

## Before → After

| Gesture | Before | After |
| --- | --- | --- |
| Double-tap the All tab | Same as a single tap | Lands on All with the search input focused, keyboard up |
| Return key on the iOS keyboard (All **and** Tasks search) | Nothing | Dismisses the keyboard, revealing the list |

## Changes

1. **`hooks/use-double-tap.ts`** (new) — the Daily tab's inline double-tap detection generalizes into a generic `useDoubleTap<Key>` hook: two taps of the same key within 450ms pair into a capture gesture, and a pending tap expires when the active key moves off it between taps. Nothing tab-specific lives in it, so it sits in the shared `hooks/` directory.
2. **`mobile-tab-bar.tsx`** — gains `tabRootFor(route)`, the route→tab mapping (which also replaces the shell's ternary chain for the highlighted tab). In the tab bar's terms, the expiry rule reads: a deep link or opened note between two tab taps means the second tap is a return, not a capture gesture.
3. **`mobile-shell.tsx`** — `handleTabSelect` composes the two (`useDoubleTap<MobileTab>(tabRootFor(route))`); a double-tap on All navigates with the router's existing `focusEditor` arrival intent, exactly like the Daily double-tap.
4. **`use-arrival-focus.ts`** (new) — a small hook that focuses a target element on `focusEditor` arrivals. It mirrors `useDailyArrivals`' bookkeeping: each `arrivalSeq` is consumed once, and the arrival that mounts the screen still counts, so a double-tap completing before the remounting screen first commits isn't swallowed.
5. **`search-input.tsx`** (new) — a shared `SearchInput` that bundles the mobile search-field config (`type="search"`, `inputMode="search"`, `text-base`) and dismisses the keyboard on the return key. The All tab (wired to `useArrivalFocus` via a ref) and the Tasks tab both use it, so the return-key behavior is identical on both — the Tasks search previously had a plain `Input` with no dismissal, which is what the reported "keyboard won't go away" screenshot was hitting.
6. **`router.tsx`** — doc-only: `focusEditor` / `arrivalFocusEditor` comments now name the All-tab search input as a consumer.

## Tests

- `use-double-tap.test.ts` — pairing rules: same-key taps within/past the window, a different key restarting the pairing, the first tap surviving the active-key change it itself causes, a mid-gesture active-key change expiring the pending tap, and the custom window parameter.
- `mobile-tab-bar.test.ts` — `tabRootFor`'s route→tab mapping (stacked screens are tabless).
- `use-arrival-focus.test.ts` — the arrival contract: a plain mount never focuses, a focus arrival that raced the mount still focuses, a capture re-arrival focuses, each arrival is consumed once, non-focus arrivals are ignored.
- `search-input.test.tsx` — return-key blurs the field, other keys don't, a caller's `onKeyDown` still runs, and the field is search-typed.

## Verification

- `pnpm check` (typecheck + lint) — clean.
- `pnpm --filter @reflect/desktop test --run` on the touched suites (`use-double-tap`, `mobile-tab-bar`, `use-arrival-focus`, `search-input`, `all-notes`, `tasks`, `mobile-screen`, `router`) — all passing.
- Live in the browser mobile harness (`?platform=ios`, 375×812): double-tap All focuses the search input; single tap does not; the Daily double-tap still focuses the editor; and the return key blurs the focused field on **both** the All and Tasks searches.

## Risk / Rollout

Mobile-only, no data or schema changes. The keyboard-raise-on-programmatic-focus and blur-to-dismiss behaviors were verified in the browser harness, not on device — worth a quick device/simulator pass before release.


maccman:
The double-tap logic nitpick is addressed: the gesture was extracted into a generic `useDoubleTap<Key>` hook (`apps/desktop/src/hooks/use-double-tap.ts`), and `use-double-tap.test.ts` pins per-key tap tracking, the 450ms window, and active-key invalidation; `mobile-tab-bar.test.ts` pins the route→tab mapping the shell feeds it.

PR #693, opened 2026-07-09
fix(editor): preserve tasks in cross-note drags

maccman:
## Problem

Each day in the daily stream owns a separate Meowdown/ProseMirror editor. ProseKit's six-dot block handle exports rendered block HTML, but not the `data-pm-slice` envelope that a different editor needs to recognize it as native content. Dropping a round task such as `+ [ ] Task` onto another day therefore ran the task DOM through Meowdown's foreign-HTML converter and produced a literal `\[ ]` paragraph followed by detached task text. The saved note no longer contained or indexed a task.

This change preserves the existing drag semantics: reordering inside one editor is still a move, while a drop between editor views is still a copy. It fixes the reported content corruption without adding cross-note source deletion.

## Before -> After

| Flow | Before | After |
| --- | --- | --- |
| Drag a round task to another daily note | Checkbox becomes `\[ ]`; task text becomes a separate paragraph | Exact task structure and `+` marker survive |
| Drag a block within one note | Reorders through ProseMirror's source slice | Unchanged |
| Drag files or other non-handle content | Existing handlers own the event | Unchanged |

## Changes

1. Add `BlockDragClipboard` inside the editor's ProseKit context. Its bubbling `dragstart` listener runs after ProseKit has selected the block, reuses ProseMirror's transformed slice metadata, and writes lossless HTML plus Markdown `text/plain` to the transfer.
2. Serialize the slice through the full Meowdown schema instead of the flat-list clipboard serializer. This retains Reflect-specific list attributes such as the round `+` marker; the normal native-list serializer would normalize it to a square `- [ ]` checklist.
3. Clear the source view's stale drag slice on `dragend`. The handle lives outside the contenteditable, so ProseMirror's own cleanup does not see cross-editor or canceled handle drags.
4. Add focused coverage using the real Meowdown schema. The tests reparse the transferred HTML, prove that `+ [ ] Task` survives exactly, isolate unrelated drags, and cover delayed mount plus listener cleanup.

## Verification

- `pnpm --filter @reflect/desktop test --run src/editor/block-drag-clipboard.test.tsx src/editor/note-editor.test.tsx` — 38 tests passed.
- `pnpm check` — typecheck and lint passed. The existing max-lines warnings in `graph-provider.tsx` and `indexer.ts` remain unchanged.

## Risk / Rollout

No migration or rollout step is required. The listener is scoped to ProseKit's block-handle custom element, and the regression covers the main isolation boundary. Cross-editor drops intentionally remain copies; implementing true cross-note moves would require destination acknowledgement before safely deleting the source.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -25,6 +25,7 @@
     "@dnd-kit/utilities": "^3.2.2",
     "@meowdown/core": "^0.42.0",
     "@meowdown/react": "^0.42.0",
+    "@prosekit/pm": "^0.1.19-beta.2",
     "@reflect/core": "workspace:*",
     "@reflect/db": "workspace:*",
     "@reflect/design-system": "workspace:*",
--- a/apps/desktop/src/editor/block-drag-clipboard.test.tsx
+++ b/apps/desktop/src/editor/block-drag-clipboard.test.tsx
@@ -1,0 +1,140 @@
+import { createElement } from 'react'
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom/vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { docToMarkdown, markdownToDoc } from '@meowdown/core'
+import { DOMParser, Fragment, Slice, type Schema } from '@prosekit/pm/model'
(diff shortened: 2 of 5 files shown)


ocavue (changes requested):
Let me re-implement this in the meowdown side.

PR #732, opened 2026-07-10
fix: preserve note aliases in links and backlinks

maccman:
## Problem

Reflect derives v1-style aliases from titles such as `Tim MacCaw // Dad`, but the surrounding link behavior was inconsistent:

- selecting `Dad` in `[[` autocomplete discarded the chosen display alias;
- autocomplete could serialize a title that opened a different note, or became ambiguous after the cross-device title-resolution hardening;
- raw suggestions, Create queries, and contact names containing wikilink delimiters could produce malformed or changed targets;
- backlinks exposed every claimant for a folded key while read navigation chose one winner.

For example, if one note is titled `Dad` and another is titled `Tim MacCaw // Dad`, bare `[[Dad]]` should backlink to the titled note, while explicitly selecting Tim's alias should insert `[[Tim MacCaw // Dad|Dad]]` and still open Tim's note.

## Before → after

| Scenario | Before | After |
| --- | --- | --- |
| Select `Dad` for `Tim MacCaw // Dad` | Loses the selected alias text | Inserts `[[Tim MacCaw // Dad\|Dad]]` |
| Select an ordinary duplicate-title row | May insert a link that writable navigation refuses as ambiguous | Omits the ambiguous address; a uniquely claimed alias can rescue the note |
| Select a valid ISO-date collision | Insertion and navigation policy can drift | Keeps deterministic date resolution and offers only the `note_keys` winner |
| Insert `[`, `]`, `\|`, `\\`, CR, or LF | Can corrupt or change the parsed wikilink | Omits the unrepresentable row instead of sanitizing it into another key |
| One key has several claimants | One link can appear in several backlink panels | One daily/title/alias winner owns read resolution and backlinks |

### Resolved address projection

- Migration 0018 changes `note_keys` from “all claims” to one winning path per folded key.
- Precedence is calendar-valid daily date → title → alias, then graph-relative path as the deterministic read/backlink tie-break.
- `claim_count` records the number of distinct notes in the winning tier. Writable non-date navigation and autocomplete use it to refuse ambiguity without changing deterministic read-only/date behavior.
- Impossible daily paths such as `daily/2026-02-31.md` remain ordinary title claimants.
- `backlinks` joins through the winner map, duplicate same-note aliases collapse, and deleting a winner promotes the next claimant automatically.
- The migration adds only a view replacement and an index; it does not rebuild or delete projection rows.

### Truthful editor suggestions

- Keeps `suggestWikiTargets()` for path-based navigation surfaces.
- Adds `suggestWikiLinkTargets()` and `WikiLinkSuggestion` for editor rows with validated `insertText`.
- Ranks the full bounded candidate set, then batches keys through `note_keys` so filtering cannot leave the menu under-filled.
- A non-date insertion is offered only when its winning resolution tier has one claimant and opens the selected path.
- Valid ISO-date targets retain the deterministic date-click contract.
- Prefers `canonical|alias`; if the canonical title is ambiguous or lost, another uniquely claimed alias can rescue the note.
- Rehydrates generated dates when the winner is the matching real daily, including custom-title dailies absent from title search.
- Returns claimed folded keys from the bounded raw candidate set before ranking/dedup, so filtered ambiguous rows still suppress misleading Create and Apple Contact actions.
- Returns date-phrase metadata from the same pass, avoiding duplicate date generation in the editor.

### Serialization boundary

- Adds `serializeWikiSuggestionAddress()` and makes `wikiSuggestionInsertText()` consume already-verified insertion text.
- Rejects `[`, `]`, `|`, backslash, CR, and LF because wikilink syntax cannot preserve them exactly.
- Applies the same rule to raw Create rows and Apple Contact names in the editor; non-Markdown consumers remain unaffected.
- Date-phrase detection uses the raw query, keeping Create suppressed even when verification drops a generated date candidate.

### Resolution contracts and documentation

- `resolveWikiTarget()` reads the canonical `note_keys` winner used by backlinks and valid-date/read-only navigation.
- Writable non-date navigation keeps the ambiguity-preserving `findExactWikiTargetMatches()` path landed in #744; autocomplete now guarantees its inserted targets satisfy that contract.
- Porting docs now describe v1's stable-ID links, v2's textual addresses, `//` rename behavior, daily aliases, collision handling, and the distinction between read and writable navigation.

## Verification

- `pnpm --filter @reflect/core test` — 95 files, 1,317 tests passed
- `pnpm --filter @reflect/desktop test` — 221 files, 1,988 tests passed
- Targeted alias/address and editor-assembly suites — 81 tests passed
- `cargo test -p reflect-open db::tests` — 54 passed
- `cargo test -p reflect-index-schema` — 2 passed
- `pnpm --filter @reflect/db db:codegen` — passed; `NoteKeys.claimCount` regenerated
- `pnpm check` — passed; only the two pre-existing max-lines warnings
- `git diff --check origin/next...HEAD` — passed

The migrated-SQL integration test now checks the actual editor click contract for every offered existing suggestion: writable non-dates resolve uniquely through `resolveOrCreateNoteWithTitle`, while valid ISO dates resolve through the deterministic read path. Coverage includes `Tim MacCaw // Dad`, standalone-title precedence, duplicate title winner/loser cases, unique-alias rescue, calendar-valid and impossible dates, generated-date hydration, and unsafe syntax.

## External review outcomes

- Fixed the reported duplicate-title regression: an unaddressable `Roadmap` claim no longer leaves a Create row that inserts ambiguous `[[Roadmap]]`.
- Fixed the same boundary for Apple Contacts, including partial queries and aliases removed by rank dedup, so contact creation cannot make a third duplicate note.
- Replaced the core test driver with Node’s built-in `node:sqlite`; `sqlite-vec` remains only because the integration test deliberately replays the complete production migration chain, whose migrations 0002/0003 create `vec0` tables.
- Audited the remaining review notes (winner-only consumers, migration ordering, bounded query cost, serialization, and docs); none requires another pre-merge change.

## Risk and rollout

- The schema change affects wiki resolution and backlinks globally, but preserves all projection rows and has explicit v17 → v18 migration coverage.
- Ordinary duplicate non-date titles are intentionally stricter: writable navigation refuses to guess, and autocomplete omits that spelling unless a unique alias addresses the note.
- The core integration test uses built-in `node:sqlite`; its sole added test dependency is `sqlite-vec`, required to replay migrations 0002/0003 exactly.
- Latest `next` (`04865f07`, beta.12) supplies `@meowdown/core` and `@meowdown/react` `^0.46.0`, which includes the separate `//` slash-menu fix from [Meowdown #276](https://github.com/prosekit/meowdown/pull/276). The typing and alias-resolution halves therefore ship together.

## Summary by CodeRabbit

* **New Features**
  * Improved wiki-link autocomplete with more reliable title, alias, and date suggestions.
  * Suggestions now avoid unsafe or ambiguous links and ensure inserted links resolve to the intended note.
  * Added deterministic resolution when notes share titles or aliases.
* **Bug Fixes**
  * Backlinks now consistently target the correct note, including after migrations and note renames.
  * Improved handling of daily notes, aliases, duplicate titles, and invalid date queries.
* **Documentation**
  * Clarified note aliases, deep-link resolution, and porting behavior.

---

> [!NOTE]
> **High Risk**
> Global wiki resolution, backlinks, and schema v18 affect every graph; behavior is stricter on duplicate titles and claimed keys, though migration preserves projection rows and adds extensive test coverage.
> 
> **Overview**
> **Index resolution (migration 0018)** replaces the old “every claimant” `note_keys` / `backlinks` views with **one winning note path per folded key**: calendar-valid daily → title → alias, ties broken by **alphabetical path**. **`claim_count`** records how many notes share that winning tier so writable flows can refuse ambiguous non-date keys without changing deterministic read/date behavior. Rust and core tests cover alias-only backlinks, precedence, v17→v18 row preservation, and winner promotion when the leader is deleted.
> 
> **Editor / core autocomplete** switches the `[[` menu to **`suggestWikiLinkTargets`** and **`WikiLinkSuggestion.insertText`**, verified against `note_keys` (unique tier for non-dates; ISO dates stay deterministic). **`serializeWikiSuggestionAddress`** rejects unrepresentable wikilink syntax; **`buildAutocompleteEntries`** uses **`claimedTargetKeys`** and **`queryReadsAsDate`** to drop misleading Create/contact rows. **`resolveWikiTarget`** now reads the canonical `note_keys` winner in one query.
> 
> Porting docs and a migrated-SQL integration test document v1 `//` aliases and the end-to-end Dad/collision flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a93c10defa0df715eb8b8b8e516ff93ced6a4bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/indexing/queries-suggestions.ts
```diff
@@ -101,7 +139,78 @@ export async function suggestWikiTargets(
...
+  for (const candidate of candidates) {
+    const canonicalKey = normalizeWikiTarget(candidate.target).key
+    if (candidate.path === null) {
+      const insertText =
+        winners.has(canonicalKey)
+          ? null
+          : serializeWikiSuggestionAddress(candidate.target, candidate.alias)
+      if (insertText !== null) {
+        verified.push({ ...candidate, insertText })
+      }
```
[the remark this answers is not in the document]
maccman: Fixed in 9862c636. Address verification now joins the resolved `note_keys` winner to `notes.daily_date`: a pathless generated date is hydrated with the existing daily path only when that winner is the matching real daily. A regular title that owns the same date-shaped key is still omitted rather than presented as a daily. Added migrated-schema coverage for both a custom-H1 existing daily (`today`) and a regular date-title collision (`tomorrow`); the full core suite (1,220 tests) and `pnpm check` pass.

packages/core/src/indexing/subject-alias-flow.test.ts
```diff
@@ -0,0 +1,403 @@
+import { readFileSync, readdirSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import Database from 'better-sqlite3'
+import { load as loadSqliteVec } from 'sqlite-vec'
```
ocavue: Why do we need `sqlite-vec` in this PR?

packages/core/src/indexing/subject-alias-flow.test.ts
```diff
@@ -0,0 +1,403 @@
+import { readFileSync, readdirSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import Database from 'better-sqlite3'
```
ocavue: Could we test this with `node:sqlite` instead of `better-sqlite3`? Notice that we can bump the CI github actions environment from nodejs v22 to nodejs v24 if we need.

ocavue:
Update: the repo `reflect-open` now only has one long-lived branch `master`. All open pull requests should target `master` instead of `next` now.

ocavue:
implemented in https://github.com/team-reflect/reflect-open/pull/861

PR #824, opened 2026-07-15
fix: sort numbered graph-switch shortcuts alphabetically

### LINKED ISSUE #816: Bug: Command+1 and Command+2 shortcuts to switch between graphs don't seem to work properly
I have two graphs and whichever one I'm in is always command+2 which is not what's documented. Plus it's awkward that since you're automatically landed in the editor for a note after you switch, if you then try to switch again using the exact same shortcuts you end up accidentally changing the heading format of a line in the note you landed in.

mvanhorn:
## Summary
Command+1 through Command+9 now switch to graphs sorted alphabetically by graph name instead of following the most-recently-used ordering, so a given number maps to a predictable graph.

## Why this matters
#816 reported that Command+1 / Command+2 graph switching behaved unpredictably. The shortcuts indexed into the graph list in MRU order, so every switch reordered the list and the number-to-graph mapping changed on each use. Pressing Command+2 twice could land on two different graphs.

## What changed
- Numbered graph slots are now ordered alphabetically by graph name (case-insensitive, root path as tiebreak) via `sortGraphsByName` in `app-commands.ts`, applied in `app-shortcuts.ts`.
- Active-graph and out-of-range selections remain no-ops.
- Regression coverage for the alphabetical ordering (`app-commands.test.ts`).

## Testing
Targeted Vitest for the changed command module: 36 passed. Typecheck + oxlint clean.

## Note
There is a separate, pre-existing collision between the Meta+digit graph shortcuts and the Meowdown heading shortcuts. That is out of scope for this fix and left unchanged; happy to file or address it separately if useful.

Closes #816

### DIFF
--- a/apps/desktop/src/lib/commands/app-commands.test.ts
+++ b/apps/desktop/src/lib/commands/app-commands.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest'
-import type { EmbedStatus, NoteRow, PinnedNote } from '@reflect/core'
+import type { EmbedStatus, NoteRow, PinnedNote, RecentGraph } from '@reflect/core'
 import { notePathForRoute, type Route } from '@/routing/route'
 import type { NavigateOptions } from '@/routing/router'
 import { resetOperations } from '@/lib/operations'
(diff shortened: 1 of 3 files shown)


maccman:
Shall we just sort graphs alphabetically?

mvanhorn:
Good call - switched to sorting alphabetically by graph name in 166f923, so Cmd+1 maps to the alphabetically-first graph and so on. More predictable than first-seen order, and it let me drop the stable-order reconciliation entirely (net simpler). Tests + typecheck green. I'll update the PR title/description to match the alphabetical approach.

ocavue:
Update: the repo `reflect-open` now only has one long-lived branch `master`. All open pull requests should target `master` instead of `next` now.

mvanhorn:
Thanks for the heads-up @ocavue — retargeted this PR to `master`.

PR #846, opened 2026-07-16
fix: prevent macOS iCloud launch crashes

maccman:
## Problem

Reflect Beta 0.7.0-beta.3 aborts (`SIGABRT`, uncaught `NSRangeException`) roughly ten seconds after launch while an iCloud-backed graph is open.

The crash report pins the cause: the iCloud watcher hands `NSMetadataQuery` a default **concurrent** `NSOperationQueue`, and CloudDocs (`BRQuery`) schedules its own internal, unsynchronized gatherer work on that client queue — not just notification delivery. The report shows three operations from the same queue executing simultaneously (the initial gather's `_classifyItems:` racing an update batch's `_classifyItems:` and `_handleReplacedItemsNotifications:`), corrupting BRQuery's result index sets until `-[NSMutableIndexSet addIndexesInRange:]` throws. The launch+10s timing is exactly the window where the initial gather overlaps the first batched update rounds. Serializing the queue closes the race.

While investigating, we also found a **separate, latent signing defect**: the macOS release helper re-signs the app with only the committed entitlement plist, which drops `com.apple.application-identifier` and `com.apple.developer.team-identifier` — identity entitlements Apple requires for Developer ID apps using iCloud, sourced from the embedded provisioning profile. The crash report shows CloudDocs was in fact accepting the broken-signature beta (the query was live when it crashed), so this is not what caused the abort — but the installed beta's signature is verifiably missing the identity entitlements, and this PR fixes and guards that too. The dev flavor also inherited the stable provisioning profile through Tauri's config merge, despite intentionally disabling iCloud entitlements.

## Before -> After

| Surface | Before | After |
| --- | --- | --- |
| iCloud metadata watch | Default concurrent operation queue | Explicitly serial queue before query startup |
| Stable/beta re-signing | Static app entitlements only | Static entitlements plus the matching embedded profile's app/team identity |
| Dev bundle | Inherited stable provisioning profile | No provisioning profile |
| Release verification | Generic code-signature validity | Also compares signed identity entitlements with the actual embedded profile |

## Changes

1. **The crash fix:** set the iCloud metadata query's `NSOperationQueue.maxConcurrentOperationCount` to `1`, serializing CloudDocs query processing and notification delivery on the queue it demonstrably shares.
2. Decode the provisioning profile embedded in each built stable/beta app, validate that its application identifier exactly matches the resolved flavor bundle ID, and merge only the profile-owned application/team identifiers into the configured entitlement plist.
3. Reject missing, conflicting, or wrong-flavor profile identity values before re-signing. Wildcard grants such as keychain access groups are deliberately not copied.
4. Verify the finalized signature retained both identity values from the actual embedded profile.
5. Remove the stable profile from the dev overlay using Tauri's merge-patch deletion semantics and document the release behavior.

## Tests

- Added release-script coverage for stable/beta/dev profile resolution, selective identity merging, exact bundle-ID matching, missing/conflicting identities, and final-signature validation.
- The existing iCloud watcher tests are pure-Rust snapshot/diff tests — they confirm the serial-queue change compiles but do **not** exercise the Objective-C queue or the concurrency itself. A manual soak is owed before calling the crash fixed: open an iCloud graph with a few thousand notes on a fresh machine so the initial gather overlaps update rounds (the exact window this crash lived in) and confirm no abort.

## Verification

- `pnpm --filter @reflect/desktop test --run scripts/release-macos.test.mjs` — 38 tests passed.
- `pnpm check` — typecheck, Oxlint, and ESLint passed (two pre-existing max-lines warnings).
- `pnpm --filter @reflect/desktop sidecar` — required desktop sidecars staged successfully.
- `cargo test -p reflect-open --lib icloud::watch` — 7 tests passed.
- `cargo fmt --package reflect-open -- --check` — passed.
- `git diff --check` — passed.
- Decoded both committed Developer ID profiles through the production `security cms` / `plutil` path and verified their stable/beta identities.
- Confirmed the new verifier rejects the currently installed broken beta signature.
- Signed and verified a temporary beta app with the local Developer ID certificate; the repaired application/team entitlements survived code signing.

## Risk / Rollout

There is no data migration or markdown change. The native behavior change only serializes metadata-query work on the queue CloudDocs already schedules onto; the cost is that gather-round enumeration now blocks update delivery briefly, which the existing 2-second batching interval already dominates. The signing change affects the next stable/beta artifacts; the release verifier now fails the build if a profile and flavor diverge or if code signing drops either identity entitlement.

## Note

This relands #842, which was merged into `next` after that branch was retired — the fix never reached `master`. This is a clean cherry-pick of the squash commit (`3d0eae96`) onto `master`; the release-script tests (38), `icloud::watch` tests (7), and `cargo fmt --check` were re-run green on the `master` base.

---

> [!NOTE]
> **Medium Risk**
> Changes native iCloud threading and the macOS release codesign path; mistakes could affect iCloud stability or block/ship broken beta/stable builds, but logic is guarded by new tests and pre-sign checks.
> 
> **Overview**
> Fixes a **~10s post-launch crash** when an iCloud-backed graph is open by forcing the `NSMetadataQuery` delivery queue to **serial** (`maxConcurrentOperationCount = 1`), so CloudDocs gather/update work cannot run concurrently on the client queue and corrupt BRQuery index sets.
> 
> **macOS release signing** now reads each stable/beta app’s **embedded provisioning profile**, validates its application identifier against the flavor bundle ID, merges only **`com.apple.application-identifier`** and **`com.apple.developer.team-identifier`** into the entitlement plist used for re-signing (not wildcard profile grants), and **fails verify** if the signed app dropped those values. **Reflect Dev** explicitly clears `embedded.provisionprofile` so it no longer inherits the stable profile.
> 
> Adds **release-script unit tests** for profile resolution, merge/reject rules, and post-sign checks; **docs** describe the updated re-sign/verify flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65f11773944e824eeb1838a2d79a182cbd87906a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS signing and verification to ensure app entitlements match the embedded provisioning profile.
  * Prevented signing mismatches that could cause release verification or installation failures.
  * Improved iCloud file monitoring stability by preventing concurrent index updates that could cause crashes.

* **Documentation**
  * Updated macOS distribution guidance to reflect the latest signing, notarization, and verification process.

### DIFF
--- a/apps/desktop/scripts/release-macos.mjs
+++ b/apps/desktop/scripts/release-macos.mjs
@@ -54,6 +54,10 @@ const INTEL_ONNX_RUNTIME_RESOURCE_SOURCE = `resources/onnxruntime/${ONNX_RUNTIME
 const RELEASE_NOTES_FILENAME = 'release-notes.md'
 const MAC_DOWNLOAD_NOTICE_HEADING = '## Which Mac download should I choose?'
 const MACOS_SIDECARS = ['reflect', 'reflect-capture-host']
+const MACOS_PROFILE_IDENTITY_ENTITLEMENTS = [
+  'com.apple.application-identifier',
+  'com.apple.developer.team-identifier',
+]
 const RELEASE_VERSION_PATTERN = /^(\d+)\.(\d+)\.(\d+)(?:-beta(?:\.(\d+))?)?$/
 const NOTARIZATION_ENV_VARS = [
   'APPLE_ID',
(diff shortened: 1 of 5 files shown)


Serenity-02:
Really solid root-causing here. The overlap of the initial gather's _classifyItems: with a batched update's _classifyItems: + _handleReplacedItemsNotifications: on the same concurrent queue explains the ~10s timing nicely — that's not a coincidence, that's exactly when the two phases collide.
Serializing the queue (maxConcurrentOperationCount = 1) makes sense as the fix since we don't control BRQuery internals. Maybe drop a comment in the code explaining why it has to stay serial, so nobody "optimizes" it back to concurrent later.
Good catch on the signing issue too, and thanks for being clear it's not what caused the crash. Since it's unrelated to the race, might be worth splitting into its own PR — easier to review/revert independently. Also, is there now a CI check so the dev flavor can't silently inherit stable's provisioning profile again? That one snuck in quietly once already.

PR #863, opened 2026-07-16
fix: keep backlink context expanded when the source item is collapsed

### LINKED ISSUE #832: The parents of incoming backlinks are not unfolded to show the context
Let's say you have
```
- Parent
    * A context about [[Child]].
```

When parent is folded
And you view the incoming backlink section inside `[[Child]]`, you only see
`+ Parent`

Should see
```
- Parent
    * A context about [[Child]].
```

RIDAM7:
## Summary

Fixes #832.

Reflect encodes a list item's fold state directly in its bullet marker:
a bullet item starting with `+` is collapsed, `-`/`*` are expanded (see
`@meowdown/core`'s `md-to-pm.ts`: `collapsed = !isTask && kind === 'bullet'
&& marker === '+'`).

`block-context.ts` builds the incoming-backlinks panel's context by slicing
the relevant lines straight from the note's source. When the item carrying
the context (or its parent, for a nested mention) is folded in the source
note, its `+` marker rides along verbatim into the returned snippet text.
`BacklinkSnippet` renders that text through the same `MarkdownView` used for
live notes, which correctly folds anything starting with `+` — hiding the
very child branch this module already went out of its way to include, even
though the returned string does contain it.

## Fix

Added `isCollapsedListItem` + `withExpandedMarker` in `block-context.ts`.
Before returning a context whose leading line belongs to a list item, if
that item's own marker is a collapsed bullet, its first display line is
normalized from `+` to `-`. This only touches the display line
(`ContextLines.lines`) — the tracked `sourceLines` used for write-back
(e.g. clicking a task checkbox in a snippet) still match the real source,
so nothing about the source note itself is touched.

Scope note: this normalizes the outermost item's own marker in each
returned context (matching the reported case and the attached test). It
does not recursively normalize a `+` on some deeper descendant folded
inside the same snippet — happy to extend it if that's wanted too.

## Test plan

- Added a regression test in `block-context.test.ts` reproducing the exact
  issue: a collapsed parent (`+ parent line`) whose nested child mentions
  the target now returns `- parent line\n  - mention of [[Target]]` instead
  of leaking the `+`.
- `pnpm --filter @reflect/core test` — 1320 tests, all passing except 2
  pre-existing failures unrelated to this change (`fts5` SQLite module
  missing in this environment; confirmed present on unmodified `master`
  too).
- `tsc --noEmit` — clean.
- `oxlint` on the changed files — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backlink context now displays collapsed `+` list markers as standard `-` bullets.
  * Preserved indentation and surrounding parent and child content in nested list contexts.
  * Improved consistency for direct and nested list-item backlink snippets.

### DIFF
--- a/packages/core/src/indexing/block-context.test.ts
+++ b/packages/core/src/indexing/block-context.test.ts
@@ -115,6 +115,25 @@ describe('blockContextAt', () => {
     )
   })
 
+  // Repro for https://github.com/team-reflect/reflect-open/issues/832
+  // A `+` bullet marker means the item is collapsed (see meowdown's
+  // md-to-pm.ts: collapsed = kind === 'bullet' && marker === '+'). The parent
+  // line is sliced from source verbatim, so its collapsed marker rides along
+  // into the snippet text. BacklinkSnippet renders that text through the same
+  // MarkdownView used for live notes, which folds anything starting with `+`
+  // — hiding the very child branch this function just went out of its way to
+  // include, even though the returned string still contains it.
+  it('does not carry the parent line collapsed marker into the context', () => {
+    const content = [
+      '+ parent line',
+      '  - mention of [[Target]]',
+      '',
+    ].join('\n')
+    expect(blockContextAt(content, posOf(content, '[[Target]]'))).toBe(
+      '- parent line\n  - mention of [[Target]]',
+    )
+  })
+
   it('keeps sibling branches that mention the same target', () => {
     const content = [
       '- parent line',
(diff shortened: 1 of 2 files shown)


ocavue:
Hi. Could you attach some screenshots below and after the fix? Thanks.

RIDAM7:
Sure! Both screenshots show the Child note's incoming-backlinks panel, with Parent folded in the source note in both cases.

**Before fix:**
BEFORE_1
BEFORE_2


**After fix:**
AFTER_1
AFTER_2

ocavue:
Thanks for your contribution. I fixed in https://github.com/team-reflect/reflect-open/pull/906 and released as v0.7.0-beta.14.

RIDAM7:
@ocavue i wanted it to show on my profile, that my raised pr got merged for this project.
sadge

PR #883, opened 2026-07-18
feat(desktop): auto-timestamp new daily-note bullets

liukun:
## Heads up

**This is a personal-workflow feature and doesn't have to be merged.** I use daily notes for time-tracking / event streaming and got tired of typing timestamps by hand. Opening a PR mostly as a record and for discussion — if maintainers feel it's too niche, feel free to close.

## Summary

Prefix a top-level bullet line with the current time (`` `HH:mm` ``) the moment its paragraph transitions from empty to non-empty. Fires uniformly for keystrokes, paste, and IME committed text. Only inside daily notes; gated by a new `editorDailyNoteAutoTimestamp` setting (default on).

## Behavior

- Fires only inside daily notes (the extension isn't mounted otherwise).
- Fires only on top-level bullet-list items (`kind: 'bullet'`). Task lists, ordered lists, nested lists, headings, code blocks, and plain paragraphs are left untouched.
- Skips lines whose first character is already a backtick — the user is typing their own code span or their own timestamp, and either way the plugin stays out of the way. This same guard also makes redo idempotent.
- Format is 24-hour `HH:mm`. Since the timestamp is written into the markdown as persistent data, 24-hour is unambiguous, sortable, and consistent across users; it deliberately does not follow `settings.timeFormat`.
- Toggleable via the new `editorDailyNoteAutoTimestamp` setting (default on) in the Editor section on desktop and mobile.

## Implementation notes

- ProseMirror plugin (`daily-note-auto-timestamp-plugin.ts`): uses `appendTransaction` to detect the empty→non-empty transition and append the stamp within the same state-advance cycle (so it stays in one undo event). A meta flag prevents self-retrigger; `mapPositionThrough` with `assoc: -1` guards against cross-block shifts; the backtick prefix check makes the redo path idempotent.
- React wrapper (`daily-note-auto-timestamp-extension.tsx`): uses `useState` lazy init (not `useMemo`) so React can never drop the memo and swap the plugin instance mid-session. The parent (`note-pane.tsx`) conditionally mounts on `dailyNote && settings.editorDailyNoteAutoTimestamp`.
- Added `@prosekit/core@^0.13.0-beta.6` as a direct dep because `@meowdown/react` re-exports `useExtension` but not `definePlugin` — that's the only stable path to hand a raw PM plugin to a ProseKit-based editor. Version range matches the one `@meowdown/react@0.55.0` already hard-pins, to avoid two prosekit-core copies with incompatible `Extension` types.

## Known limitations

- Host-initiated `insertMarkdown` (attach files, insert template) into an empty bullet also fabricates a stamp — the plugin can't tell "user typed" from "host inserted" without an out-of-band signal. Debatable whether that's a bug or a feature.
- A forward-`Delete` that merges a next non-empty bullet into an empty one triggers the same transition and gets stamped. Same root cause.
- 24-hour format does not follow `settings.timeFormat`. Trade-off explained above.

## Test plan

- [x] Unit tests cover 9 scenarios: first-character stamps, subsequent typing doesn't restamp, IME committed text, backtick-first skip, non-empty paragraph skip, clear-and-retype restamps, task-list skip, ordered-list skip, plain-paragraph skip.
- [x] Desktop `typecheck` and `vitest` green.
- [ ] Manual smoke: type in daily note, paste multi-line, commit via CJK IME, toggle setting off/on, open non-daily note (should not stamp).

## Summary by CodeRabbit

* **New Features**
  * Added a setting to automatically add timestamps when starting entries in daily-note bullet lists.
  * Enabled the feature by default, with controls available in desktop and mobile editor settings.
  * Timestamps are added only to eligible new entries and avoid duplicate or unintended formatting.

* **Tests**
  * Added coverage for typing scenarios, list types, clearing and re-entering content, and timestamp persistence.


maccman:
Sorry, this one isn't relevant for most people.

PR #889, opened 2026-07-18
fix: move a task instead of duplicating it when dragged between notes

### LINKED ISSUE #747: Drag and drop a task from one note to another should not duplicate the task item.
https://github.com/user-attachments/assets/ec30ba95-8cc3-4a7c-89e5-d55ce67e0235

RIDAM7:
Fixes #747

## Problem
The daily stream mounts one ProseMirror `EditorView` per day, so dragging a task
from one day onto another is a drag **between two editor instances**. ProseMirror
only deletes dragged content in the view where the drop lands, so the target
inserted a copy while the source kept the original — the task ended up in both notes.

## Fix
The block-handle grip is a floating element outside any `view.dom`, so this adds a
**document-level** bridge (`cross-note-drag-move.tsx`) that tracks the mounted note
views, and on the source view's `dragend` — when the drop landed in a *different*
editor and it was a move — runs the delete ProseMirror skipped, using PM's own drag
record (`dragging.node.replace`, the same call a same-view move uses). A cancelled
drop or a copy drag deletes nothing, so neither can lose the original.

The move decision is factored into a pure `crossNoteMoveDeletion` helper with unit
tests. Native drag-and-drop can't run under vitest/jsdom, so the gesture itself was
verified manually in the running app; the automated coverage is the decision logic.

## Testing
- `pnpm typecheck`, `pnpm lint`, `pnpm test` pass
- Manually verified in `pnpm tauri:dev`: dragging a task between daily notes now
  moves it to a single note; a drop cancelled mid-drag leaves the original intact.

## Summary by CodeRabbit

* **New Features**
  * Added support for moving content between note editors using drag and drop.
  * Source content is automatically removed after a successful cross-note move.
  * Copy operations, cancelled drops, same-editor moves, and empty selections remain unchanged.

* **Tests**
  * Added coverage for cross-editor moves, copies, cancellations, and selection handling.

### DIFF
--- a/apps/desktop/src/editor/cross-note-drag-move.test.ts
+++ b/apps/desktop/src/editor/cross-note-drag-move.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+import { crossNoteMoveDeletion } from './cross-note-drag-move'
+
+/** The settled state of a task/block move dropped in another day's editor. */
+const crossNoteNodeMove = {
+  isSourceView: true,
+  droppedInAnotherView: true,
+  move: true,
+  hasNode: true,
+  selectionEmpty: true,
+}
+
+describe('crossNoteMoveDeletion', () => {
+  it('deletes the dragged node when a move lands in another editor (issue #747)', () => {
+    expect(crossNoteMoveDeletion(crossNoteNodeMove)).toBe('node')
+  })
+
+  it('deletes the source selection for a cross-editor text-selection move', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


RIDAM7:
This is Screen Recording after fix #747.

https://github.com/user-attachments/assets/120be6f1-8055-4174-917e-d51bf943b388

RIDAM7:
@ocavue pls review

ocavue:
Hi. This has been fixed in 0.7.0-beta.19 with an alternative implementation so I'm going to close this PR. Thanks for your contribution!

PR #896, opened 2026-07-20
fix: handle iCloud dataless files

maccman:
## Problem

natan's report: hours after importing a V1 graph into an iCloud directory, Reflect freezes and `fileproviderd` pins at 230% CPU; force-closing calms it, reopening brings it back. A main-thread sample on the affected machine showed 99–100% of samples in `note_read → read_to_string → read`, with most notes being **dataless** files (nonzero size, zero blocks).

Three compounding defects:

1. **Modern macOS eviction was invisible.** All placeholder detection knew only the legacy `.name.md.icloud` stub grammar (the iOS form). Since macOS 12.3, iCloud Drive evicts to *dataless files at the real path*, so `list_files` reported evicted notes as ordinary readable files and every eviction guard (indexer skip, reconcile-scan exclusion, pending counts) never fired. Any read of such a file blocks while `fileproviderd` materializes it.
2. **Whole-graph passes read evicted notes serially on the main thread.** The per-open embedding backfill reads every note (the hash-skip avoids re-*embedding*, never re-*reading*); `note_read` is a sync Tauri command, so each blocking download ran on the UI thread. That's the freeze, and why reopening re-triggers it.
3. **The metadata watch fought eviction, unpaced.** The gather round requested a download for every placeholder on every watch start, and deliberately re-nudged after each eviction — a tug-of-war with Optimize Mac Storage that keeps `fileproviderd` pegged as long as the app runs.

## Fix

- **Dataless detection** via the kernel `SF_DATALESS` flag (`st_flags`), folded into the one `placeholder` flag in `collect_files`, `pending_walk`, and the FSEvents watcher's evicted-not-deleted rule. Deliberately not `st_blocks == 0`: decmpfs-compressed files also allocate zero blocks and would be silently dropped from indexing forever.
- **`note_read_local`** — a new async (off-main-thread) command returning content-or-`evicted` without materializing. The embedding backfill and asset-description gathering skip evicted files; pre-eviction vectors and index rows stay valid, so search keeps working.
- **Changed-only nudges** in the metadata watch: request a download only for content the device lacks — a new arrival, or a remote edit reaching an evicted item (the snapshot now tracks placeholder content-change dates). Never the gather round, never an eviction of already-seen content. iOS keeps its open/resume `icloud_download_pending` ladder as the bulk driver.
- **Targeted reconcile downloads**: `index_reconcile_scan` now reports placeholders whose row is missing or stale; the desktop requests exactly those via the new `icloud_request_downloads` (traversal-guarded, resolves against the active root). Rebuilds surface *all* placeholders, so a projection-bump wipe can't silently drop evicted notes from search — they re-index via ordinary watcher upserts as they materialize.

Net behavior: an OS-evicted graph stays evicted (Optimize Mac Storage is respected), the index serves pre-eviction content, opening a note still materializes it on demand, and only content the index actually lacks is downloaded.

## Testing

- `cargo test -p reflect-open` — 310 passed (new coverage: dataless flag decode, scan stale-placeholder classification, changed-only nudge policy incl. eviction-stays-silent and remote-edit-renudge).
- `pnpm --filter @reflect/core test --run` — 1406 passed (new: evicted note skipped without touching stored vectors; evicted sidecar skipped; reconcile/rebuild surface stale placeholders).
- `pnpm --filter @reflect/desktop test --run` — 2043 passed; `pnpm check` clean.
- **Manual pass owed** (needs a real iCloud account): `brctl evict` a few graph notes, then verify the listing flags them, open causes no nudge storm, the backfill skips them, and opening an evicted note in the editor still materializes it. Two-device check that a remote edit to an evicted note downloads and re-indexes.

## Notes / residual risk

- A read racing a fresh eviction can still materialize one file (stat-then-read) — harmless; the pass-level storm is what's prevented.
- Legacy stub placeholders list the stub's own mtime, so on iOS an evicted-after-indexing note re-requests per reconcile — matching mobile's existing download-all-pending behavior.
- `note_read` itself is still sync; moving the remaining blocking fs commands off the main thread is a follow-up PR.

---

> [!NOTE]
> **Medium Risk**
> Touches iCloud sync, indexing reconcile, and file I/O paths that affect large-graph open behavior; logic is heavily tested but real iCloud eviction still needs manual verification.
> 
> **Overview**
> Fixes freezes and `fileproviderd` storms when a graph lives in iCloud by treating **modern macOS dataless files** (kernel `SF_DATALESS`) the same as legacy `.icloud` stubs everywhere listing, watching, and pending walks decide a note is present but not local.
> 
> **Reads and background work** no longer force materialization: new async `note_read_local` returns content or `evicted`; embedding backfill and asset-description gathering use it and skip evicted notes/sidecars without wiping stored vectors or doing a full chunk replace when a sidecar is missing.
> 
> **Download policy** stops fighting Optimize Storage: the metadata watch skips nudges on the gather round and only requests downloads for provably missing content (new arrivals, remote edits on evicted items with comparable content dates). Reconcile adds `stale_placeholders` for evicted paths whose index row is missing or stale; the desktop calls `icloud_request_downloads` for exactly those paths while leaving evicted notes with current rows alone.
> 
> **Watcher / FSEvents** treat dataless files like evictions (no upsert that would trigger blocking reads). `pending_walk` counts and nudges both stub and dataless forms.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ae2debffaa954d18be0d54e5653692de34a0af1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added iCloud “targeted downloads” support via a new request command, and surfaced `stalePlaceholders` during index rebuild/reconcile to drive these requests.
  * Added local-only note reads that distinguish evicted notes (`{ evicted }`) without forcing materialization.
  * Asset description and embedding now skip evicted sidecar content.
* **Bug Fixes**
  * Improved iCloud evicted-file detection (including modern dataless evictions) and prevented watcher events from treating them as real content.
* **Tests**
  * Expanded reconciliation, iCloud eviction, embedding, and asset-description tests for the new `stalePlaceholders` and local-read behaviors.


ocavue:
Addressed the latest CodeRabbit round: `9938404c` qualifies the `note_read_local` doc comment as best-effort (matching the call-site caveat), and `7bd3fdbd` makes the `AssetDescriptionGather` fields `readonly` per the coding guidelines. The Copilot review generated no comments.

ocavue (approved):

PR #342, opened 2026-06-23
Add Reflect / Reflect Beta / Reflect Dev build flavors

ocavue:
Ships three coexisting macOS apps via Tauri `--config` overlays: **Reflect** (stable, purple, `app.reflect.desktop`), **Reflect Beta** (`next`, red, `app.reflect.desktop.beta`), and **Reflect Dev** (local, yellow, `app.reflect.desktop.dev`).

`release:macos` derives the flavor from the version (prerelease → beta, else stable), so `release.yml` and `release:bump` are unchanged.

| Flavor | Branch | productName | identifier | Icon | Updater feed |
| --- | --- | --- | --- | --- | --- |
| Reflect | `master` | Reflect | `app.reflect.desktop` | purple | `releases/latest` |
| Reflect Beta | `next` | Reflect Beta | `app.reflect.desktop.beta` | red | `updater-beta` |
| Reflect Dev | local | Reflect Dev | `app.reflect.desktop.dev` | yellow | none |

Local: `pnpm tauri:dev` (Reflect Dev), `pnpm tauri:build:beta`. Docs in `docs/macos-distribution.md` → Build flavors.

### Notes for review
- GitHub rewrites spaces in asset names to dots, so the beta updater URL is sanitized to `Reflect.Beta.app.tar.gz` in `writeUpdaterManifest` (else auto-update 404s).
- Distinct identifiers isolate webview storage + embeddings; settings, recent graphs and keychain secrets stay **shared** (hardcoded `reflect-open`). Intentional for now.
- **Beta migration:** the first flavored beta is a new app id, so existing beta testers must delete the old "Reflect" beta and install "Reflect Beta" once. Stable installs are unaffected (icon just turns purple).

### Before un-drafting
- [ ] Visually confirm the yellow Dev / red Beta / purple stable icons (hue picked numerically)
- [ ] Run a real local build per flavor (`pnpm tauri:build:dev` / `:beta`) and confirm all three coexist in /Applications
- [ ] Sanity-check a signed beta (`pnpm release:macos --flavor=beta --no-notarize`)

PR #899, opened 2026-07-20
fix: add Paste and Match Style (⌘⇧V) to the macOS Edit menu

maccman:
## Problem

⌘⇧V (paste without formatting) does nothing in the mac app. The app uses Tauri's default macOS menu, which has no "Paste and Match Style" item — and WKWebView only performs editing shortcuts that an app-menu key equivalent routes to it, so the keystroke reaches the webview as a bare keydown with no paste behind it. The old Electron app bound the same accelerator to `pasteAndMatchStyle:` (reflect-electron `src/menu.ts`), which is why this worked in traditional Reflect.

## Changes

**Rust (`src-tauri`)**
- New `menu.rs` (macOS-only): builds Tauri's default menu and inserts "Paste and Match Style" (`CmdOrCtrl+Shift+V`) directly under Edit > Paste. The default Edit submenu has no stable id, so it's matched by title with a warn-and-degrade fallback to the stock menu.
- The click handler reads the pasteboard via `tauri-plugin-clipboard-manager` (registered macOS-only, used Rust-side only — no webview capability grant) and `emit_to`s the focused window with the text (`menu:paste-and-match-style`).

**Frontend**
- `@reflect/core`: `PASTE_AND_MATCH_STYLE_EVENT` + `subscribePasteAndMatchStyle` in `app/window-events.ts` (zod-validated payload, same shape as `subscribeWindowNavigate`).
- New `PasteAndMatchStyleBridge`, mounted inside every `NoteEditor`'s ProseKit context (like `EditorInputTraits`): the focused editor runs the text through `view.pasteText` — ProseMirror's plain-text paste path, the exact one a browser Shift-paste takes. Covers note panes, the daily stream, and the task editor; a no-op without a native shell.

## Why not `editor.commands.pastePlainText`

meowdown [PR #342](https://github.com/prosekit/meowdown/pull/342) adds a `pastePlainText` command plus an in-editor ⌘⇧V clipboard-read fallback, but it isn't released yet (and this repo pins `@meowdown/*` 0.46). `view.pasteText` is the same code path that command wraps and exists in every meowdown version, so this fix doesn't need the bump. When meowdown is next bumped past 0.55.1, its in-editor fallback also becomes a second layer of coverage.

## Testing

- New `paste-and-match-style-bridge.test.tsx`: pastes on the event, ignores it when unfocused, rejects malformed payloads, unsubscribes on unmount, no-ops without a bridge.
- `pnpm check` (typecheck + lint) green; desktop editor tests green; `cargo check`, `cargo clippy`, and `cargo test -p reflect-open` green.
- Not yet verified in a built app bundle — reviewers should confirm in `pnpm tauri:dev`: ⌘⇧V pastes clipboard text without formatting, plain ⌘V still pastes with formatting, and the Edit menu shows the item under Paste.

---

> [!NOTE]
> **Low Risk**
> macOS-only menu/clipboard wiring with graceful fallback if the Edit submenu is missing; no auth or data-model changes, and clipboard is read only on explicit user menu/shortcut.
> 
> **Overview**
> Restores **paste without formatting** on macOS by wiring ⌘⇧V through the native app menu instead of relying on WKWebView, which never received a menu key equivalent for that shortcut.
> 
> On **Rust (macOS only)**, the shell registers `tauri-plugin-clipboard-manager`, builds Tauri’s default menu with a new **Edit → Paste and Match Style** item (`CmdOrCtrl+Shift+V`) under Paste, reads plain text from the pasteboard on activation, and `emit_to`s the focused window on `menu:paste-and-match-style`. Clipboard access stays in the shell (no webview capability).
> 
> **`@reflect/core`** adds `PASTE_AND_MATCH_STYLE_EVENT` and zod-validated `subscribePasteAndMatchStyle`. **`PasteAndMatchStyleBridge`** listens in each `NoteEditor` and calls `view.pasteText` only when that editor is focused; tests cover focus, bad payloads, unsubscribe, and no bridge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46e5b76582426d32e3b3f1039b20eb75ac9aec8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,6 +145,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "arboard"
+version = "3.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0348a1c054491f4bfe6ab86a7b6ab1e44e45d899005de92f58b3df180b36ddaf"
+dependencies = [
+ "clipboard-win",
+ "image",
+ "log",
+ "objc2",
+ "objc2-app-kit",
+ "objc2-core-foundation",
+ "objc2-core-graphics",
+ "objc2-foundation",
+ "parking_lot",
+ "percent-encoding",
+ "windows-sys 0.60.2",
+ "wl-clipboard-rs",
+ "x11rb",
+]
+
+[[package]]
 name = "arg_enum_proc_macro"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -741,6 +762,15 @@ name = "clap_lex"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
+
+[[package]]
+name = "clipboard-win"
+version = "5.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bde03770d3df201d4fb868f2c9c59e66a3e4e2bd06692a0fe701e7103c7e84d4"
+dependencies = [
+ "error-code",
+]
 
 [[package]]
 name = "color_quant"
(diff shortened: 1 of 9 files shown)


ocavue:
Replaced by https://github.com/team-reflect/reflect-open/pull/905, which is simpler because it uses the native WebKit API instead of calling ProseMirror

PR #901, opened 2026-07-21
fix: create daily notes from date links

jasonlong:
## Summary
- materialize a missing daily note when an ISO or relative-date wiki-link suggestion is selected
- use the atomic no-clobber note creation boundary so concurrent/existing daily files are never replaced
- surface background creation failures instead of silently leaving a dangling link
- cover exact dates, relative dates, existing dailies, invalid dates, and failures

## Context
The backlink symptom came from date autocomplete inserting the ISO link without running the V1-compatible daily creation side effect. This fixes that creation path directly rather than synthesizing backlinks for notes that do not exist.

## Verification
- `pnpm check`
- `pnpm build`
- `pnpm --filter @reflect/core test src/graph/create-note.test.ts`
- `pnpm --filter @reflect/desktop test src/editor/use-editor-autocomplete.test.tsx`

## Summary by CodeRabbit

- **New Features**
  - Daily notes are now automatically materialized from date-based autocomplete selections, including “pathless date” entries.
  - Newly inserted daily-note links resolve immediately and gain backlinks without requiring navigation.
- **Bug Fixes**
  - Existing daily notes are preserved and not recreated.
  - Invalid dates are rejected before any write attempts.
  - Daily-note creation failures now surface operation errors with the underlying cause.


maccman:
Can you help us understand a bit more on the context here? We should be auto-creating daily notes automatically when you type a date in or a relative time like "three days from now".

jasonlong:
Ugh, for some reason my LLM is commenting on my behalf here. In a nutshell, the original issue I was seeing was this:

1. Start typing `[[next Mon` in today's daily note and select an autocomplete option
2. ISO formatted link is properly inserted into note
3. Click on link for next Monday and there's no backlink from today

Leaving LLM's comment below...

----

You’re right — I traced this back through the V1 backlink-menu contract. Selecting an ISO or generated relative-date suggestion is supposed to materialize the daily note immediately. The current V2 autocomplete inserted the normalized ISO target, but unlike the ordinary Create row, the pathless date suggestion had no `onSelect` creation callback.

I initially addressed the downstream backlink symptom in the schema, which was the wrong layer and would have made manually typed dangling date links look like backlinks to nonexistent notes. I’ve replaced that approach in dc5827da: selecting a new date suggestion now atomically creates the empty `daily/YYYY-MM-DD.md` file (without replacing an existing/concurrently-created daily), and both exact and relative date paths are covered by tests. Once indexed, the existing backlinks view works unchanged.

PR #914, opened 2026-07-23
feat: show backlinks on daily notes that have no file yet

Jeminaa:
## Problem

Linking a future date (e.g. `[[2026-07-29]]` as a reminder) stores the link in the index immediately, but when that day arrives the daily note's backlinks panel is empty until something is typed. Daily notes are created lazily on the first keystroke, and backlink resolution goes through the `backlinks` view, whose `note_keys` join requires an indexed target note — so a daily that exists only as a link target has no rows to serve. Past dates appear to work only because those files already exist.

## Change

`getBacklinks` / `getBacklinksWithContext` now resolve over the `links` table with the target's full key set — its `note_keys` rows plus, for a daily path, the date derived from the path itself (`dateFromDailyPath`). The view's semantics are preserved (wiki links only, template sources excluded, alias/title spellings still co-group), and the same key set feeds the block-context highlighter, replacing the second `note_keys` query.

A future daily now lists its inbound links the moment the date is opened, before any keystroke creates the file.

## Testing

- New `queries-backlinks.test.ts`: absent-daily resolution through the path-derived date key, note_keys + date key merging, template-source exclusion, empty-key short circuit, page-limit guard, and links-table (not view) query shape.
- `pnpm check` clean; verified in a real graph (link from one daily to a not-yet-created future daily shows up in the panel).

https://claude.ai/code/session_01GLgoFLw2b4xv1cLudoyPLT

### DIFF
--- a/packages/core/src/indexing/queries-backlinks.test.ts
+++ b/packages/core/src/indexing/queries-backlinks.test.ts
@@ -1,0 +1,116 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { setBridge } from '../ipc/bridge'
+import { getBacklinks, getBacklinksWithContext } from './queries-backlinks'
+
+const mockInvoke = vi.fn<(command: string, args: Record<string, unknown>) => Promise<unknown>>()
+
+beforeEach(() => {
+  mockInvoke.mockReset()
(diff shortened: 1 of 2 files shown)


Jeminaa:
Apologies — opened prematurely, withdrawing.

PR #940, opened 2026-07-23
fix: keep automatic daily entries in their lists

maccman:
## Problem

Automatic daily-note entries were appended at the end of an entire category section. If a `## [[Links]]`, `## [[Audio memos]]`, or `## Meetings` section began with bullets and later contained prose or another block, the next automatic entry landed below that unrelated content instead of extending the list.

A nested example heading inside a quote or list could also be mistaken for the real category section. Worse, when the only matching heading was nested, the entry was appended at end of file with no heading at all.

Two related defects the old end-of-section append also carried:

- **Mixed line endings.** `source.slice(0, sectionEnd).replace(/\s*$/, '')` trimmed a trailing `\r\n` and reattached a bare `\n`, so every automatic write flipped one line of a CRLF note to LF. Plan 03 asks edits not to reflow a file, or sync drowns in spurious diffs.
- **Loose lists.** Appending at the section end produced `- [[Old]]\n\n- [[New]]`. That is one list, but a *loose* one, and meowdown's serializer normalizes it back to `- [[Old]]\n- [[New]]` the next time the user edits that note. The old output therefore guaranteed a diff unrelated to anything the user did. The new output is what the editor would have written.

## Before -> After

| Daily-note shape | Before | After |
| --- | --- | --- |
| Heading, leading bullets, then prose | New entry appended after the prose | New entry extends the bullet list |
| Heading, prose, then a bullet list | New entry appended at the section end | New entry extends the existing list |
| Heading followed only by prose | New entry appended at the section end | A new bullet starts directly beneath the heading |
| Heading, prose, then a subheading with a list | New entry appended at the section end | A new bullet starts beneath the heading, never inside the subsection |
| Matching heading nested in a quote/list | Bullet appended at EOF with no heading | Preserved as authored; a top-level category section is created |
| CRLF note | Edited line silently became LF | Line endings preserved |

## Changes

1. Add syntax-tree-aware list insertion (`appendListItemAtHeading`) that targets the section's first top-level `BulletList`, stopping at any heading so an entry never lands in a subsection.
2. Preserve existing list markers (switching the bullet character would split one list into two), frontmatter offsets, local CRLF line endings, and later authored content.
3. Use the new insertion contract for captured links, audio memos, and calendar meetings. The contract takes the item's **content**; the bullet marker belongs to the list being extended, so callers no longer pass one.
4. Align capture and meeting deduplication with the same top-level heading boundary, while continuing to recognize older misplaced links anywhere in a valid section. Both dedup scans are deliberately wider than where a new entry lands, and now say so.
5. Centralize the pieces this touched: `node-types.ts` keys node checks on meowdown's `LEZER_NODE_IDS` so a `@lezer/markdown` rename is a compile error rather than a regex that silently stops matching, and `line-endings.ts` replaces three separate CRLF detectors. `Heading` carries `topLevel`, recorded during `parseNote`'s single walk, so `topLevelHeadings` no longer reparses the body.
6. Drop `appendUnderHeading`, which lost its last caller.
7. Add focused regressions for list/prose boundaries, missing lists, ordered lists, subsection lists, renamed and legacy backlinked headings, nested headings, CRLF, frontmatter, marker reuse, captures, audio memos, and meetings.

## Tests

- Captured links extend the Links list even when the section opens with the user's own prose, and ignore nested category examples.
- Audio memo backlinks stay before later prose.
- Meetings use the same list boundary on the current attendee-resolution flow.
- Markdown helpers cover missing lists, ordered lists, subsection lists, nested lists, later blocks, aliases, legacy headings, frontmatter, marker reuse, and CRLF.

## Verification

- `pnpm test --run packages/core`: 101 files, 1477 tests passed.
- `vitest run --project node`: 75 files, 702 tests passed.
- `pnpm check`: passed (typecheck and lint; the three `max-lines` warnings are pre-existing on files this PR does not touch).
- Behaviour checked directly against `origin/master` for every shape in the table above.

## Risk / Rollout

No migration or backfill is required; `Heading.topLevel` is in-memory only and never persisted, so the index needs no rebuild. Existing misplaced entries are left untouched; future automatic entries use the corrected boundary. The change is limited to Markdown insertion and section-deduplication behavior.


maccman:
Addressed both CodeRabbit nits in f6d3ee74: section-boundary lookup now uses the shared `sectionEnd` helper, and `appendBlock` / `appendUnderHeading` preserve CRLF. Added focused CRLF regressions; 173 targeted tests and `pnpm check` pass.

maccman:
Addressed the follow-up CodeRabbit documentation nit in 8eeebabd by documenting the top-level heading helper’s source-offset and frontmatter behavior. The focused 79-test Markdown suite and `pnpm check` pass.

ocavue:
### Replies to the CodeRabbit review threads

All three nits were addressed before this push; here is where each one stands after it.

**1. "Section-boundary lookup is duplicated across three files"** (`append-section.ts`): **fixed**, and still fixed.

`sectionEnd` lives in `heading-blocks.ts` and is the only implementation. `meetingAlreadyLinked` (`add-meeting.ts`) and `findSameDayCapture` (`capture-drain.ts`) both call it. The third copy, inside `appendUnderHeading`, is gone with the function itself (see 2).

**2. "`appendBlock` / `appendUnderHeading` don't preserve CRLF"**: **fixed**, and hardened.

`appendBlock` and `appendHeadingSection` now get their line ending from a single `documentLineEnding` in the new `markdown/line-endings.ts`, which replaced the three separate CRLF detectors that had accumulated in this package (`insertionLineEnding` in `edit.ts`, `documentLineEnding` in `append-section.ts`, `nearbyLineEnding` in `append-list-item.ts`). `upgradeSectionHeadingBacklink` uses the same module's `offsetBeforeLineEnding` instead of its own inline `\r` check.

`appendUnderHeading` is deleted rather than fixed: after this PR moved the meetings flow onto `appendListItemUnderHeading`, it had no production caller left, only a barrel export and its own tests.

**3. "Document the exported `topLevelHeadings` API"**: **fixed**, and the doc was rewritten here.

The old doc described source-offset and frontmatter behaviour because the function reparsed the body to recover which headings were top-level blocks. It no longer does: `parseNote` records `topLevel` on each `Heading` during the walk it was already doing, and `topLevelHeadings` is now a filter over that flag. The JSDoc says what the flag means and why a heading inside a blockquote or list item is not a section.

ocavue:
### Review response

I went through this branch in detail, comparing behaviour against `origin/master` on the same inputs rather than reading the diff alone. The direction is right: the bug is real, and a syntax-tree boundary is the correct fix rather than scanning the source with regexes. Below is what I changed and what I deliberately left alone. `origin/master` is merged in.

#### Fixed

**Insertion targeted the block flush against the heading, which split lists and reversed chronology.** This was the one blocking issue. Given a section that opens with the user's own prose and *then* holds the category list:

```md
## [[Links]]

Reading list for today:

- [[Old capture]]
```

three consecutive captures produced a second list above the prose and stranded the first one:

```md
## [[Links]]

- [[New 1]]
- [[New 2]]
- [[New 3]]

Reading list for today:

- [[Old capture]]
```

The same shape appears in this PR's own updated test `upgrades the existing plain Links section without duplicating it`, where the new capture now sorts above the older one, while `orders same-day captures chronologically` asserts the opposite for the ordinary shape.

`appendListItemAtHeading` now takes the section's **first top-level `BulletList`**, scanning forward from the heading and stopping at any heading of any level. That keeps the guarantee the PR wanted (an entry never lands inside a subsection, never crosses into another section) and drops the one it did not intend. Regressions added for the prose-first-then-list shape, an ordered list, and a list belonging to a subheading.

**Stringly-typed node checks.** `/^(?:ATX|Setext)Heading[1-6]$/` was the third spelling of this predicate in `packages/core/src/markdown` (`extract.ts` had `/^(?:ATXHeading|SetextHeading)([1-6])$/`, `wiki-nodes.ts` uses `satisfies LezerNodeName`), and it has no failure signal: if `@lezer/markdown` renames a node, the regex silently stops matching and every automatic write quietly degrades to "section not found, create a new one".

meowdown already exports what this needs. New `markdown/node-types.ts` keys the predicates on `LEZER_NODE_IDS`, which is `Readonly<Record<LezerNodeName, number>>` derived from `gfmParser` itself and pinned against `gfmParser.nodeSet` by a unit test upstream. A rename is now a compile error. `extract.ts`'s copy is gone too. Two details worth recording: `SetextHeading3`..`6` do not exist (setext is only `===` and `---`), so the `[1-6]` branch was partly dead, and `node.type.is('Heading')` is **not** usable here, because `@lezer/markdown` only attaches that group to custom-defined nodes, not to its built-in heading types.

**Three line-ending detectors.** Merged into `markdown/line-endings.ts`; see the CodeRabbit reply above.

**`topLevelHeadings` reparsed the document.** It took `parseNote`'s `headings` and then parsed the body again purely to recover which of them were direct document blocks, information the first walk had and discarded. `Heading` now carries `topLevel`, set during that walk. Parses per call:

| | master | before this push | now |
| --- | --- | --- | --- |
| `appendListItemUnderHeading` | 1 | 3 | 2 |
| `appendListItemUnderBacklinkedHeading` | 2 | 5 | 3 |

The remaining extra parse is `appendListItemAtHeading`'s own, which genuinely needs block nodes rather than headings. `Heading` is never persisted (only a `heading TEXT` scalar on embedding chunks, derived at chunk time), so this needs no `PARSED_NOTE_VERSION` bump and no index rebuild.

**Runtime `throw` for a shape the type system should carry.** `withListMarker` required callers to pass `- text`, validated the prefix at runtime, then rewrote it to the list's actual marker. The throw was unreachable: all three callers hard-code the prefix. The bullet marker belongs to the list being extended, not the caller, so `appendListItem*` now takes the item's content and `withListMarker` / `normalizeListItem` are gone. `meetingLine` returns the text without the marker; the rendered daily line is unchanged.

**Dedup scans the whole section while insertion touches one list.** On master both were the whole section, so they agreed; narrowing insertion made them disagree without saying so. The asymmetry is right (a meeting a user dragged below their notes is still on the day's agenda, and an entry an older build misplaced should not be captured twice), so it is now stated in the docstrings of `meetingAlreadyLinked` and `findSameDayCapture` rather than left to be read as an oversight.

**`appendUnderHeading` had no callers left.** Deleted, along with its tests and its two barrel exports.

**Smaller things.** `capture-drain.ts` import order restored; the "a list item target must be a top-level heading" guard now actually checks the block is a heading instead of only that some top-level block starts at that offset.

**PR description.** Rewritten to include two effects the original did not mention, both of which are the strongest justification for touching this code at all: the old append flipped one line of a CRLF note to LF on every write, and it emitted a *loose* list that meowdown's serializer rewrites to a tight one on the user's next edit, guaranteeing a sync diff nobody asked for. I confirmed the round trip in meowdown: `- a\n\n- b\n` serializes back as `- a\n- b\n`, and `*` / `+` markers are preserved, which is also why marker reuse here is correctness rather than polish (changing the bullet character starts a new list in CommonMark).

#### Not changed

**`if (!target)` -> `if (target === undefined)`.** Kept. It is behaviourally identical (`Heading` is an interface, always truthy) and it does match the two sibling functions that already used `=== undefined` before this PR, so the family now reads uniformly. Worth flagging only because it landed inside an otherwise pure move of the function into a new file, which makes the move harder to verify at a glance; reverting it now would be more churn than leaving it.

**A section whose only content is a non-list paragraph still gets a new list under the heading**, so a new entry sits above a legacy bare `[[capture-…|Old]]` line. That is the intended "heading followed only by prose" row of the table, and the legacy shape is only produced by an older build. Recorded here so it is a decision rather than an accident.

**`## Meetings` is the only category heading that is not a backlink**, unlike `## [[Links]]` and `## [[Audio memos]]`. That is deliberate v1 compatibility per `docs/porting/calendar-meetings-integration.md`, and out of scope here, but now that all three flows share one insertion API it would be a cheap thing to reconsider separately.

#### Correction to the automated summaries

The CodeRabbit and Cursor summaries both say duplicate detection "now respects link aliases". It already did: `meetingAlreadyLinked` on master matches `foldKey(link.alias)` and carries a comment explaining why. Same for `an un-backlinked meeting always appends` and `still appends when the title is only linked outside the Meetings section`. Those three tests are welcome, but they pin existing behaviour rather than introduce it, so there is no alias behaviour change to review.

#### Verification

- `pnpm test --run packages/core`: 101 files, 1477 tests passed.
- `vitest run --project node`: 75 files, 702 tests passed.
- `pnpm check`: passed. The three `max-lines` warnings are pre-existing on files this PR does not touch.
- Every "Before -> After" row in the description was executed against both `origin/master` and this branch rather than reasoned about.

PR #960, opened 2026-07-26
fix: V1 import no longer fails on graphs with 1000+ same-named attachments

maccman:
## Problem

A user importing a multi-year Reflect V1 graph hit a hard import failure:

> **Import failed** — no free asset name after 1000 probes for image.png

Every image pasted into Reflect V1 uploads under the clipboard's generic filename, `image.png`, and Firebase serves that name back via `Content-Disposition`. The import names downloaded attachments after that original filename, and its collision probing only tried sequential suffixes: `image.png`, `image-2.png`, … `image-1000.png`. A graph with more than 1000 distinct pasted screenshots exhausts the cap on the 1001st one, and `plan_asset_name` aborts the entire import — deterministically, so retries and restarts can never help (which is exactly what the reporter saw).

The same 1000-probe cliff exists in `persist_unique`, the intake path behind in-app paste/drop uploads and file-picker imports, so a graph dense with one filename would eventually break ordinary pasting too.

## Before → After

| Scenario | Before | After |
| --- | --- | --- |
| V1 import with ≤ ~8 same-named attachments | `image.png`, `image-2.png`, … | Unchanged |
| V1 import with 1000+ distinct `image.png` attachments | Whole import aborts with "no free asset name after 1000 probes" | Each asset lands as `image-<8-hex-sha256>.png` after the readable suffixes fill up |
| Re-importing the same export | Reuses files found during sequential probing | Same, and digest names re-derive deterministically from content, so reuse still short-circuits |
| Paste into a graph dense with one stem | Up to 1000 rename syscalls, then a hard error | Jumps to a digest candidate after 8 sequential probes |

## Changes

1. **`fs/assets.rs`** — new `NameCandidates` prober shared by both intake paths. Candidate sequence: the desired name, then readable `-2`…`-8` suffixes (`SEQUENTIAL_NAME_PROBES = 8`), then names carrying the first 8 hex chars of the file's sha256 (`image-3f9ab2c1.png`, `image-3f9ab2c1-2.png`, …). The digest is computed lazily — the file is only read once sequential probing runs dry — and cached per sequence. `MAX_NAME_PROBES = 1000` remains as a backstop, but digest candidates are effectively unique per distinct file, so real graphs resolve in a couple of probes. `persist_unique` now iterates `NameCandidates`; its `persist_noclobber` claim semantics (collision check *and* claim, race-free) are unchanged.
2. **`fs/import_assets.rs`** — `plan_asset_name` drops its private copy of the sequential loop and cap and iterates the shared `NameCandidates` instead. Determinism matters here: planning the same bytes again re-derives the same digest name, which is how a re-import finds the file it wrote last time and marks it `reuse` instead of duplicating it. The `taken`-set check, on-disk existence/`file_occupied` checks, and identical-bytes reuse logic are unchanged.

Digest names are content-derived, not random, on purpose: import retries and re-imports stay idempotent, and two distinct files whose 8-hex prefixes ever collided would simply fall through to the `-2` variant of the digest name.

## Tests

- `plan_switches_to_digest_names_when_sequential_probes_exhaust` (`import_assets.rs`) — with `image.png` through `image-8.png` all taken, the plan lands on an 8-hex digest name; the same bytes re-plan to the same name with `reuse: true`; different bytes land beside it under their own digest.
- `persist_switches_to_digest_names_when_sequential_probes_exhaust` (`assets.rs`) — the upload path falls over to a digest name once the sequential window is dense on disk, and the persisted bytes are intact.
- Existing probing/reuse tests (`plan_reuses_identical_bytes_and_probes_conflicts`, `persist_probes_numbered_suffixes_on_collision`, …) pass unchanged, pinning that behavior below the 8-suffix window is identical to before.

## Verification

- `cargo test --lib` in `apps/desktop/src-tauri`: 330 passed, 0 failed.
- `cargo clippy --lib`: clean.

## Risk / Rollout

- No migration and no change to existing files on disk — only the names newly persisted assets receive once a stem already has 8+ collisions.
- Behavior for the common case (first collision on a name) is byte-for-byte identical, so existing graphs see no difference.
- A partial import made by a previously failed run is safe to re-run: identical bytes found on disk (or planned in-run) are reused, not duplicated.

---

> [!NOTE]
> **Low Risk**
> Scoped to new asset filenames after heavy collisions; first eight suffixes and noclobber/reuse semantics are unchanged, with no migration of existing files.
> 
> **Overview**
> Fixes **hard import/upload failures** when a graph has more than ~1000 distinct files sharing one stem (e.g. thousands of V1 `image.png` pastes). Collision handling no longer stops at `image-1000.png`; after **eight** readable `-2`…`-8` suffixes it uses **deterministic** names from the first 8 hex chars of the file’s SHA256 (`image-3f9ab2c1.png`, with `-2` variants if needed).
> 
> **`fs/assets.rs`** introduces shared `NameCandidates` (lazy hash, `SEQUENTIAL_NAME_PROBES = 8`) and wires **`persist_unique`** (paste/drop and file-picker intake) through it.
> 
> **`fs/import_assets.rs`** drops its duplicate probe loop; **`plan_asset_name`** uses the same prober, adds **`legacy_numbered_reuse_name`** so re-imports can still reuse older `image-9.png`-style files after the digest switch, and **`same_file_bytes`** compares large files incrementally instead of loading them whole.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3be517a4c6c78f5be429880708656397a94b3bce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Improvements**
  * Asset imports now use readable numeric suffixes (`-2` through `-8`) for initial filename conflicts.
  * When those suffixes are unavailable, assets receive deterministic, content-based names.
  * Re-importing identical files consistently reuses the existing asset, including legacy numbered filenames.
  * Different files receive distinct names, with reliable handling for large files through incremental processing.
* **Bug Fixes**
  * Improved handling of filename conflicts during asset persistence and import.


maccman:
Do not merge this until I've done more investigation.

ocavue:
@maccman is this one ready for review?

PR #963, opened 2026-07-26
fix: cut iCloud sync's battery cost on large graphs

maccman:
## Problem

The iOS battery audit found the two iCloud maintenance passes scale badly with graph size:

1. **`pending_walk` (the placeholder count/nudge) walked the entire graph with no exclusions.** It recursed into `.git/` and `.reflect/` — thousands of loose git objects on a backup-repo graph — paying two `stat`s per entry (`file_type` for the symlink check, then a discarded `path.is_dir()`), up to 20 times per foreground while downloads drained, on a flat 1 Hz poll with no visibility gate. Both directories are marked sync-excluded at bootstrap (`NSURLUbiquitousItemIsExcludedFromSyncKey`), so descending into them is provably zero-yield: iCloud can never hold a placeholder there.
2. **Every conflict sweep ran one `NSFileVersion` check per non-placeholder note** — each a synchronous file-coordination round-trip to `fileproviderd` — even though the `NSMetadataQuery` watch already reads `HasUnresolvedConflicts` per item and emits the exact conflicted paths. During a bulk sync that meant a full O(N)-XPC pass every 30-second arrival window.

## Changes

**Rust — `icloud/storage.rs`**
- `pending_walk` never enters sync-excluded directories (shared `local_only_name` predicate, renamed from `adopt_skips` — same set, same reason) and reuses the readdir-record `file_type` for the dir check, halving per-entry stats.

**Rust — `icloud/watch.rs`**
- New `CONFLICTED` set maintained by the same notification rounds that emit `icloud:conflicts` (pure `apply_conflict_delta`, unit-tested beside `apply_update_delta`), gated by a `GATHERED` flag so a not-yet-seeded or stopped watch answers "unknown", never "no conflicts". Exposed as `conflicted_paths()`.

**Rust — `icloud/sweep.rs`**
- `icloud_conflicts_scan` gains `scope: 'full' | 'candidates'`. `'candidates'` restricts **only** the per-note version checks to the watch's conflicted set — ingest base advances, collision folding, and the baseline still see the full listing — and degrades to full whenever the watch can't answer. Housekeeping (`prune_orphans` + `archive::prune`) runs only on full sweeps; pruning against a narrower view would drop bases for merely-unlisted notes, and deferring it to the next resume is safe.

**Frontend — `icloud-controller.ts`**
- Scope threads through `scheduleScan` with a sticky-`'full'` merge rule: resume, watch-failed, and the adoption baseline sweep stay full (the backstop); conflict-signal and arrival sweeps pass `'candidates'` (the signal and the set come from the same notification round). A failed sweep re-arms `'full'` so the retry is thorough.

**Frontend — `use-icloud-refresh.ts`**
- The pending poll backs off geometrically (1s → 2s → 4s cap; the fast early ticks stay since notes usually land within a poll or two), bails while `document.visibilityState === 'hidden'` (matching every sibling controller), and restarts cleanly on the next resume.

## Testing

- Rust: `cargo test -p reflect-open icloud::` — 37 pass, including new tests for the walk's exclusions, the conflict-set delta semantics, scoped-sweep housekeeping deferral, and scoped sweeps still folding collision duplicates. `cargo fmt --check` + clippy clean.
- TS: `icloud-controller.test.tsx` (14 pass) with new scope-per-trigger and failed-sweep-retries-full tests; `use-icloud-refresh.test.tsx` (9 pass) with new backoff-pacing and hidden-stop/resume-restart tests. `pnpm check` clean.
- Owed: a device pass on a real iCloud graph (large graph + a second device producing conflicts) before trusting the sweep scoping under real `fileproviderd` behavior.

## Risk

The sweep scoping is the sensitive piece. Safety rests on three invariants, each covered: `None`/not-gathered degrades to a full check (Rust-side), deferred dirty-session conflicts stay provider-flagged until actually resolved so they remain candidates, and resume/baseline sweeps are always full. The stale-direction asymmetry is deliberately harmless-only: a just-resolved path may get one extra no-op check; a new conflict cannot be missed because the triggering signal and the candidate set come from the same round. The walk exclusion is zero-risk (sync-excluded dirs cannot contain placeholders). The poll backoff trades up to ~3s of extra latency on a slow drain for ~3× fewer full-graph walks.

---

## Update (2026-08-01): rebased + review round addressed

Rebased over the merged startup-perf stack (#1008–#1010): the sweep rides `blocking::run_blocking` and keeps the timing log (now with `scope`); the gather-round reconcile emit and `has_base` fill guard from #1010 coexist with the scoping unchanged.

Review fixes (eef8c622):

- **Epoch-scoped view** — readiness, authority, and paths now live in one `ConflictView` under one lock; rounds fold in only when their install epoch is current, and `conflicted_paths()` requires the epoch to still be live. A stop/restart can no longer hand a candidates sweep `Some(empty)` or let a stale gather re-seed a newer watch.
- **Query-coverage gate** — watch start records (off the main thread) whether the query's scope actually covers the root; a desktop graph outside the app container answers `None` and every candidates sweep degrades to full.
- **Queued scope** — mid-sweep requests keep their scope through the replay (full sticky), so an overlapping conflict signal no longer escalates to a full O(N) version pass. Tests: replay-scope assertions on both mid-sweep tests.

Verification re-run after rebase + fixes: `cargo test -p reflect-open` 355 pass, clippy/fmt clean, iOS target check clean; `icloud-controller.test.tsx` 15 pass, `use-icloud-refresh.test.tsx` 10 pass; `pnpm check` clean. The owed real-device pass (large graph + second device producing conflicts) still stands as the gate before trusting sweep scoping under real `fileproviderd` behavior.

## Summary by CodeRabbit

* **New Features**
  * Improved iCloud conflict detection by prioritizing likely conflict locations while retaining full-scan coverage when needed.
  * Added smarter iCloud download polling with exponential backoff.
  * Paused pending-download polling while the app is hidden and resumed it when visible again.
  * Improved iCloud directory traversal to skip excluded folders and symbolic links.

* **Bug Fixes**
  * Ensured failed targeted conflict scans retry with full coverage.
  * Prevented housekeeping cleanup during limited-scope scans.
  * Ignored stale conflict notifications to keep results accurate.

---

> [!NOTE]
> **Medium Risk**
> Scoped conflict sweeps are the main risk; safety relies on degrading to full when the watch is unknown or non-authoritative, epoch-gated conflict views, and resume/baseline always staying full—still warrants real-device validation under fileproviderd.
> 
> **Overview**
> Reduces **iCloud sync maintenance cost** on large graphs by making two hot paths cheaper and tightening when full work runs.
> 
> **Placeholder walks** (`pending_walk`) no longer recurse into sync-excluded trees (`.git/`, `.reflect/`, `.DS_Store` via shared `local_only_name`), and directory checks reuse the `readdir` `file_type` instead of an extra stat. A non-creating `ubiquity_documents_path()` supports the watch’s coverage probe without planting an empty Reflect folder in Drive.
> 
> **Conflict sweeps** gain `scope: full | candidates`. In **candidates** mode, per-note `NSFileVersion` checks run only on paths the live metadata watch flags (`ConflictView` with install epoch, gather readiness, and container-authority gating); ingest advances, collision folding, and baselines still use the full listing, and shadow/archive housekeeping runs only on **full** sweeps. Watch start probes whether the query actually covers the graph root (off the main thread); graphs outside the app ubiquity container always degrade to full.
> 
> The **desktop iCloud controller** passes candidates for conflict signals and arrival ingests, keeps **full** sticky for resume, adoption baseline, and failed-scan retries, and preserves queued scope through mid-sweep replays. **Mobile** pending-download polling uses geometric backoff (1s→4s cap) and stops while hidden, restarting on the next visible resume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c71ba677a60ccbd44b9953d7ff6ac0be442f1ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/desktop/src-tauri/src/icloud/watch.rs
+++ b/apps/desktop/src-tauri/src/icloud/watch.rs
@@ -51,6 +51,19 @@ pub fn icloud_watch_stop(app: tauri::AppHandle) -> AppResult<()> {
     platform::stop(app)
 }
 
+/// The paths the live metadata query currently flags as carrying unresolved
+/// conflict versions — the scoped conflict sweep's candidate set. `None`
+/// whenever the query cannot answer completely (no watch installed, or the
+/// gather round hasn't seeded the view yet); callers must treat `None` as
+/// "check everything". The set may run momentarily stale in the harmless
+/// direction only: a just-resolved path lingers until its update round lands
+/// (an extra no-op check), while a new conflict is always in the set before
+/// the `icloud:conflicts` signal that triggers a sweep, because both come
+/// from the same notification round.
+pub(crate) fn conflicted_paths() -> Option<std::collections::HashSet<String>> {
+    platform::conflicted_paths()
+}
+
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 mod platform {
     use std::collections::{HashMap, HashSet};
@@ -117,6 +130,20 @@ mod platform {
     static SNAPSHOT: LazyLock<Mutex<HashMap<String, TrackedState>>> =
         LazyLock::new(|| Mutex::new(HashMap::new()));
 
+    /// Paths the provider currently flags as carrying unresolved conflict
+    /// versions — maintained by the same rounds that emit `icloud:conflicts`,
+    /// and consumed by the scoped conflict sweep so it checks `NSFileVersion`
+    /// only where the provider says a conflict exists instead of once per
+    /// note. Valid only once [`GATHERED`] is set: before the gather round the
+    /// set is empty for the wrong reason.
+    static CONFLICTED: LazyLock<Mutex<HashSet<String>>> =
+        LazyLock::new(|| Mutex::new(HashSet::new()));
+
+    /// True once the current watch's gather round has seeded a complete
+    /// view. Cleared on install and teardown — a stopped or restarted watch
+    /// must answer "unknown", never "no conflicts".
+    static GATHERED: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
+
     /// Content-change date last download-requested, per graph-relative path.
     /// The OS treats repeat requests as no-ops, but *issuing* them is not
     /// free: during an initial sync every update round used to re-request
@@ -172,6 +199,9 @@ mod platform {
     /// caller is a main-thread closure, which is what serializes teardown
     /// against installs.
     fn teardown_active(mtm: MainThreadMarker) {
+        // A stopped watch must answer "unknown", never "no conflicts".
+        GATHERED.store(false, std::sync::atomic::Ordering::SeqCst);
+        CONFLICTED.lock().expect("conflict lock").clear();
         let Some(bound) = ACTIVE.lock().expect("watch lock").take() else {
             return;
         };
@@ -217,6 +247,8 @@ mod platform {
         }
         SNAPSHOT.lock().expect("snapshot lock").clear();
         NUDGED.lock().expect("nudge lock").clear();
+        CONFLICTED.lock().expect("conflict lock").clear();
+        GATHERED.store(false, Ordering::SeqCst);
         let query = NSMetadataQuery::new();
         query.setNotificationBatchingInterval(UPDATE_BATCHING_INTERVAL_S);
 
@@ -441,6 +473,35 @@ mod platform {
             .collect()
     }
 
+    /// Fold one round's items into the conflicted-path set: a reported item
+    /// enters or leaves by its flag, a removed item leaves. Pure, so the set
+    /// semantics are unit-testable next to [`apply_update_delta`]'s.
+    fn apply_conflict_delta(
+        conflicted: &mut HashSet<String>,
+        upserted: &[ItemState],
+        removed: &[String],
+    ) {
+        for item in upserted {
+            if item.conflict {
+                conflicted.insert(item.rel.clone());
+            } else {
+                conflicted.remove(&item.rel);
+            }
+        }
+        for rel in removed {
+            conflicted.remove(rel);
+        }
+    }
+
+    /// The scoped sweep's candidate set — see the outer
+    /// [`super::conflicted_paths`] for the contract.
+    pub(crate) fn conflicted_paths() -> Option<HashSet<String>> {
+        if !GATHERED.load(std::sync::atomic::Ordering::SeqCst) {
+            return None;
+        }
+        Some(CONFLICTED.lock().expect("conflict lock").clone())
+    }
+
     /// One gathering/update round. Updates apply the notification's own
     /// added/changed/removed delta — O(changed items); a full results
     /// enumeration here would be O(all items) per round, O(n²) across an
@@ -526,6 +587,10 @@ mod platform {
                 nudged.remove(rel);
             }
         }
+        {
+            let mut conflicted = CONFLICTED.lock().expect("conflict lock");
+            apply_conflict_delta(&mut conflicted, upserted, removed);
+        }
         Round {
             changes,
             conflicts: conflicted_rels(upserted),
@@ -572,6 +637,14 @@ mod platform {
                 .collect();
             apply_update_delta(&mut snapshot, &items, &removed)
         };
+        {
+            // A full listing is authoritative — rebuild the set wholesale,
+            // and only a full listing may declare the view complete.
+            let mut conflicted = CONFLICTED.lock().expect("conflict lock");
+            conflicted.clear();
+            apply_conflict_delta(&mut conflicted, &items, &[]);
+        }
+        GATHERED.store(true, std::sync::atomic::Ordering::SeqCst);
         Round {
             changes,
             conflicts: conflicted_rels(&items),
@@ -701,10 +774,10 @@ mod platform {
     #[cfg(test)]
     mod tests {
         use super::{
-            apply_update_delta, plan_nudges, root_variants, tracked_note_relpath, ItemState,
-            TrackedState,
+            apply_conflict_delta, apply_update_delta, plan_nudges, root_variants,
+            tracked_note_relpath, ItemState, TrackedState,
         };
-        use std::collections::HashMap;
+        use std::collections::{HashMap, HashSet};
 
         fn state(entries: &[(&str, TrackedState)]) -> HashMap<String, TrackedState> {
             entries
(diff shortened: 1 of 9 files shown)


apps/desktop/src/lib/icloud-controller.ts
```diff
@@ -145,10 +161,16 @@ export function createIcloudController(options: IcloudControllerOptions): Icloud
...
+   * {@link INGEST_SCAN_MIN_SPACING_MS} from the previous sweep's end. Scoped
+   * to the watch's conflict candidates — during a bulk sync these fire every
+   * spacing window, and a full per-note version check each time was the
+   * sweep's dominant cost on a large graph. */
   function scheduleIngestScan(): void {
     const sinceLastScan = Date.now() - lastScanEndedAt
-    scheduleScan(Math.max(INGEST_SCAN_DEBOUNCE_MS, INGEST_SCAN_MIN_SPACING_MS - sinceLastScan))
+    scheduleScan(
+      Math.max(INGEST_SCAN_DEBOUNCE_MS, INGEST_SCAN_MIN_SPACING_MS - sinceLastScan),
+      'candidates',
```
maccman: **[P1] Keep arrivals full when the metadata query is not authoritative**

This controller also mounts for `com~apple~CloudDocs` roots, but the file's own contract notes that the metadata query covers only the app container. On desktop, these arrival events come from the independent `notify` watcher anyway, so they do not share an update round with `CONFLICTED`. After the query gathers an empty/incomplete view, this candidate sweep can therefore skip every `NSFileVersion` check and leave a newly arrived conflict silent until the next focus/resume. Only request `candidates` when the query is authoritative for this graph; otherwise keep this path `full` (or have Rust return `None` for an uncovered root).

apps/desktop/src-tauri/src/icloud/watch.rs
```diff
@@ -441,6 +473,35 @@ mod platform {
...
+        }
+        for rel in removed {
+            conflicted.remove(rel);
+        }
+    }
+
+    /// The scoped sweep's candidate set — see the outer
+    /// [`super::conflicted_paths`] for the contract.
+    pub(crate) fn conflicted_paths() -> Option<HashSet<String>> {
+        if !GATHERED.load(std::sync::atomic::Ordering::SeqCst) {
```
maccman: **[P1] Make readiness and candidates one epoch-scoped snapshot**

`GATHERED` and `CONFLICTED` are synchronized separately, and notification callbacks are not tied back to the watch epoch. During stop/restart, a scan can read `GATHERED == true`, then clone the set after teardown cleared it, returning `Some(empty)` instead of `None`; an already-running old `full_round` can also set `GATHERED` back to true after teardown or after a new graph installs. Either race makes a candidate sweep trust an empty/stale set and skip real conflicts. Keep `{ epoch, gathered, paths }` under one lock and ignore notification rounds whose epoch is no longer current.

apps/desktop/src/lib/icloud-controller.ts
```diff
@@ -116,15 +117,30 @@ export function createIcloudController(options: IcloudControllerOptions): Icloud
...
-  function scheduleScan(delayMs: number = SCAN_DEBOUNCE_MS): void {
+  function scheduleScan(
+    delayMs: number = SCAN_DEBOUNCE_MS,
+    scope: IcloudSweepScope = 'full',
+  ): void {
     if (disposed || scanSuspended()) {
       return
     }
+    if (scope === 'full') {
+      nextScanScope = 'full' // sticky until a sweep consumes it
```
maccman, line 148: **[P2] Preserve candidate scope when replaying a queued prompt**

When a conflict signal arrives while `scanRunning`, its `candidates` scope is not stored with `queuedScan`. The `finally` block later replays every prompt via `scheduleScan()` with the default `full`, which sets this sticky flag and turns the follow-up into another O(N) version pass. That is especially likely when a large-graph sweep overlaps new notifications, precisely the case this PR is optimizing. Queue the requested scope alongside the urgency (merging `full` as sticky), and replay with that scope; add a scope assertion to the existing mid-sweep conflict test.

maccman (commented):
Found three issues in the candidate-sweep fast path: two can turn an incomplete/stale metadata view into an authoritative empty set and skip conflict checks, and one defeats the intended optimization for signals queued during a sweep.

apps/desktop/src-tauri/src/icloud/watch.rs
maccman: Fixed in eef8c622: the view is now one `ConflictView { epoch, authoritative, gathered, paths }` under a single lock. Rounds carry the epoch their watch was installed with and `fold_conflicts_into_view` drops any round whose epoch isn't the view's — a late gather from a torn-down query folds into nothing. `conflicted_paths()` reads everything in one lock acquisition and additionally requires `view.epoch == EPOCH`, which also covers the window where stop() has bumped the epoch but its main-thread teardown hasn't run. Pure tests cover the stale-round drop and the rebuild/seed semantics.

apps/desktop/src/lib/icloud-controller.ts
maccman: Fixed in eef8c622 via your second option — Rust answers None for an uncovered root. `icloud_watch_start` (now async) resolves whether the root lives inside the app's own ubiquity container off the main thread (the container API can block on first use; both sides canonicalized for the /var symlink) and records it as `authoritative` in the epoch-scoped view. A CloudDocs-root graph therefore always gets None → every candidates sweep degrades to full, on both platforms; TS keeps requesting candidates and the degrade stays in one place.

apps/desktop/src/lib/icloud-controller.ts
maccman: Fixed in eef8c622: `queuedScan` now carries `{ urgency, scope }`, merging the way scopes always merge (full sticky, prompt beats ingest), and the replay passes the stored scope through. The existing mid-sweep conflict test now asserts the replay's `candidates` scope, and a new test covers a mid-sweep resume keeping full through the merge.

apps/desktop/src/mobile/use-icloud-refresh.ts
```diff
@@ -69,15 +75,22 @@ export function useICloudRefresh(): void {
...
       retryTimer = setTimeout(() => {
         retryTimer = null
         if (disposed) {
           return
         }
+        if (document.visibilityState === 'hidden') {
+          // Backgrounded mid-drain: stop rather than keep walking the graph
+          // behind a hidden webview (every sibling controller suspends on
+          // hidden). The next resume restarts the whole sequence.
+          return
```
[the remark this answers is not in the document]
maccman: Addressed in b74ae90f (pre-existing on the branch): the hidden-bail clears the dedupe stamp so a hide → show inside the dedupe window restarts the poll; test included.

PR #995, opened 2026-07-30
feat: support transcription for openai-compatible providers

Yiipu:
## Problem

#996 intentionally scoped `openai-compatible` to chat, editor AI transforms, capture enrichment, asset descriptions, and audio memo title generation — "where the configured chat model can be used directly." Transcription was excluded because the `/v1/audio/transcriptions` endpoint is a different API surface from `/v1/chat/completions`, and the model types differ (speech-to-text vs chat).

However, for users with OpenAI-compatible transcription endpoints (local Whisper servers, Groq, or API proxies that expose `/v1/audio/transcriptions`), being able to configure transcription through the same `openai-compatible` provider entry is natural — the base URL is the same, only the model differs.

## Changes

1. **Schema**: `openai-compatible` entries gain an optional `transcriptionModel` field (empty = transcription not supported). A new `defaultTranscriptionProviderId` setting is introduced, independent of the chat default.

2. **Catalog**: `AiProviderInfo` gains `supportsTranscription` — a compile-time constant for hosted providers (OpenAI/Google: true, Anthropic/OpenRouter: false), and `false` for `openai-compatible` (the user opts in per-entry via `transcriptionModel`). `aiProviderSupportsTranscription()` merges both sources.

3. **Provider-config**: The hardcoded `TRANSCRIPTION_PROVIDERS` constant is replaced by `transcriptionProviders()` — a dynamic function that derives eligible providers from the configured state. `resolveTranscriptionTarget` and `pickTranscriptionConfig` now work with any provider that has transcription capability.

4. **Transcribe client**: `transcribeAudio` accepts optional `baseUrl` and `model` params. When the provider is `openai-compatible`, the client posts to `{baseUrl}/audio/transcriptions` with the user-supplied model.

5. **Audio memo pipeline**: `memoNoteBody` forwards `baseUrl` + `transcriptionModel` to the transcription client when the provider is `openai-compatible`.

## Still deferred

- Settings UI for the new `transcriptionModel` field and `defaultTranscriptionProviderId`

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAI-compatible AI providers with custom endpoints and models.
  * Added a separate default provider setting for audio transcription.
  * Improved provider selection for transcription, including automatic fallback options.
  * Added clearer handling for providers that require API keys and those that support transcription.

* **Bug Fixes**
  * Improved transcription error handling and API key resolution.
  * Normalized and validated custom provider URLs for safer configuration.


Yiipu:
Closing this PR in favor of #1007.

The scope grew beyond what the original PR body described. #995 started as transcription-only, but the actual diff now includes independent chat and transcription defaults, a `disabled` sentinel for openai-compatible entries, and the full desktop and mobile settings UI. The original body listed five change areas and explicitly deferred the settings UI. The bot review also landed on that stale description.

#1007 starts from the same commits with a clean review surface and an accurate PR body. It stays draft until I manually open it.

PR #1007, opened 2026-07-31
feat: support OpenAI-compatible transcription and independent chat/transcription defaults

Yiipu:
## Problem

#996 added `openai-compatible` providers for chat, editor AI transforms, capture enrichment, asset descriptions, and audio memo title generation. That change did not include transcription. The `/v1/audio/transcriptions` endpoint is separate from `/v1/chat/completions`. The model types also differ (speech-to-text versus chat).

Users with OpenAI-compatible transcription endpoints want to route transcription through the same provider entry. These endpoints include local Whisper servers, Groq, and API proxies that expose `/v1/audio/transcriptions`. The base URL stays the same. Only the model differs.

This PR supersedes #995. The scope grew beyond transcription. It now adds independent chat and transcription defaults. It also adds a `disabled` sentinel for openai-compatible entries and the full desktop and mobile settings UI.

The original PR body no longer matches the diff. The bot review also landed on a stale description. This PR starts from the same commits with a clean review surface.

### Schema and catalog

- `openai-compatible` entries gain an optional `transcriptionModel` field. An empty string (the default) and the `'disabled'` sentinel both mean the entry does not transcribe.
- A new `defaultTranscriptionProviderId` setting is independent of the chat default. Chat and transcription can use different providers.
- `AiProviderInfo` gains `supportsTranscription`. This is a compile-time constant for hosted providers. OpenAI and Google set it to true. Anthropic and OpenRouter set it to false. For `openai-compatible`, the catalog declares `false`. The user opts in per entry via `transcriptionModel`.
- `aiProviderSupportsTranscription()` merges the catalog constant with the per-entry `transcriptionModel`.
- `aiProviderSupportsChat()` lets an `openai-compatible` entry opt out of chat by setting `model` to `'disabled'` (a Whisper-only endpoint).
- The catalog offers a `DISABLED_OPENAI_COMPATIBLE_MODEL` sentinel as an option so it round-trips through the settings schema.

### Provider config

- This PR replaces the hardcoded `TRANSCRIPTION_PROVIDERS` constant with `transcriptionProviders()`. This function derives eligible providers from the configured state.
- `resolveTranscriptionTarget` and `pickTranscriptionConfig` work with any provider that has transcription capability.
- `defaultTranscriptionProvider()` resolves the transcription default the same way `defaultAiProvider()` resolves the chat default.
- `TranscriptionProvider` widens to include `'openai-compatible'`. This PR removes `TranscriptionConfig` and uses `AiProviderConfig` instead.

### Transcription client

- `transcribeAudio` accepts optional `baseUrl` and `model` params. When the provider is `openai-compatible`, the client posts to `{baseUrl}/audio/transcriptions` with the user-supplied model.
- Hosted OpenAI transcription keeps its hardcoded model and fallback retry. The fallback only runs when the caller did not supply a model.
- `httpError` reports the actual provider, not a hardcoded `'openai'`.

### Audio memo pipeline

- `reconcileAudioMemos` resolves the key through `aiApiKeyForConfig` (shared with chat) instead of a direct keychain read. This lets openai-compatible entries with an optional key work.
- `transcribeSessionParts` forwards `baseUrl` and `transcriptionModel` when the provider is `openai-compatible`.

### Chat model options

- `chatModelOptions()` filters out entries whose chat model is the `'disabled'` sentinel. The picker cannot switch to a chat-disabled entry.

### Settings UI (desktop and mobile)

- The add-provider dialog and the mobile add-provider sheet show a second model picker for transcription when the provider is `openai-compatible`. Hosted transcription-capable providers show their fixed model as a disabled combo.
- Both "Use as default for chat" and "Use as default for transcription" checkboxes appear when the provider supports the feature.
- The provider row on desktop shows a second model combo and a second default control for transcription. The mobile settings list badges both defaults.
- The mobile provider-actions drawer shows the transcription model section and the transcription default action.
- `ModelCombobox` gains a `disabled` prop for the fixed-model display.

## Screenshots

image

image

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (no new warnings)
- [x] Targeted tests pass: `provider-config`, `provider-catalog`, `transcribe`, `secrets`, `model-options`, `schema`, `audio-memo`, `ai-providers-section`, `transcription-reconciler`, `add-ai-provider-drawer`, `ai-provider-actions-drawer` (196 tests total)
- [x] No `any` types, zod at boundaries, kebab-case filenames
- [x] Public APIs documented with doc comments
- [ ] iOS UI test (blocked: the base-ui `Select` inside the add-provider `Dialog` has a pre-existing interaction bug. The dropdown flashes open then becomes unresponsive on the next click. You cannot manually verify the new transcription model picker on iOS. This PR did not introduce the bug. It affects the existing provider picker too.)

## Relation to #995

#995 is the same work at an earlier stage. This PR replaces it. The commits are identical. Only the PR body and review surface are new.

## Summary by CodeRabbit

* **New Features**
  * Configure separate chat and transcription providers and default models.
  * OpenAI-compatible providers support custom transcription models and endpoints.
  * Choose local, hosted, or disabled models through model selectors.
  * Mobile and desktop settings show transcription capabilities and defaults.
* **Bug Fixes**
  * Providers without chat support are no longer offered as chat defaults.
  * Audio memo availability and setup guidance recognize all transcription-capable models.
  * Improved handling of unavailable or explicitly selected transcription models.


Yiipu:
Closing this — I'm no longer maintaining this contribution and don't have a near-term need for the feature myself. The use case (OpenAI-compatible transcription endpoints) may still be worth picking up natively; #995 can be referenced if anyone wants to take it forward.

PR #1009, opened 2026-07-31
fix(ios): stop the brief input freeze right after the app opens

maccman:
Stacked on #1008. Second leg of the startup-perf investigation: the iOS report — "the UI boots and I can see a note, but for 500ms–1s it's frozen."

## Problem

Sync Tauri commands execute on the app's main thread. On iOS that thread also owns UIKit hit-testing and touch delivery, while the WKWebView renders in a separate process — so the UI paints normally while taps go nowhere. The repo already knew this pattern: the asset protocol was made asynchronous for exactly this reason, and `background_task.rs` documents the main-thread synchrony. The post-paint boot burst — one uninterruptible O(graph) `index_reconcile_scan` (directory walk + full `notes`-table read, slower still on an iCloud container), plus a volley of `db_query` and `note_read` calls from the three mounted day panes — all ran there. Resume was worse: the reconcile fires twice (index lifecycle + iCloud refresh, coalesced into one pass but always a full one).

## What changed

`index_reconcile_scan`, `db_query`, `note_read`, `list_files`, and `capture_shared_inbox_relay` become `async` commands running their bodies on the blocking pool — the pattern `note_read_local`, the asset protocol, and the git/iCloud commands already use. The hop itself is now one shared helper (`blocking::run_blocking`, hoisted from `git/mod.rs`), used by all of these plus the pre-existing `note_read_local` and sweep sites. Commands needing managed state inside the closure take `AppHandle<R>` and resolve it there; the others resolve the root before hopping. No frontend changes: `invoke` was already async.

Ordering is unchanged by design: the `IndexState` mutex — not the main thread — is what serializes reads against index writes (its doc comment states this as the contract), and every boot-path call site awaits sequentially.

## Deliberately excluded (wave 2)

`index_open` and the index **write** commands stay sync. They hold `ScopedBackgroundTask` assertions whose expiration safety currently depends on main-thread synchrony — UIKit delivers the expiration handler on that thread, so it cannot fire mid-command today. Moving them off-main first needs the expiration→cancel signal that `background_task.rs`'s comment demands, otherwise iOS could suspend the app mid-write while it holds a SQLite lock. That design is a tracked follow-up.

## Testing

- `cargo test -p reflect-open` — 341 passed (the two direct `db_query` test call sites now `block_on`).
- `cargo check -p reflect-open --target aarch64-apple-ios` — clean, covering the iOS-only relay half. `cargo clippy` clean.
- No TS changes; the dev bridge's command surface is name/payload-compatible.
- Real-device acceptance (tap responsiveness in the first second after paint on a large graph) still owed — same device pass the earlier iOS perf PRs are waiting on.

---

> [!NOTE]
> **High Risk**
> Touches iOS main-thread IPC, index concurrency/graph-switch semantics, and release-time native symbol upload—any mistake can cause wrong-graph data, boot regressions, or unsymbolicated production crashes.
> 
> **Overview**
> **Fixes the post-paint “frozen for ~1s” window on iOS** by turning filesystem/SQLite-heavy commands (`index_reconcile_scan`, `db_query`, `note_read`, `list_files`, `capture_shared_inbox_relay`) into async handlers that run on a shared `blocking::run_blocking` pool instead of the touch-delivery main thread. `db_query` now uses a **read-only SQLite connection** so long FTS work does not contend with sync index writes. Reconcile scans pin the **index session’s graph root** (not live `graph_open` state) and generation checks guard async races; a new test covers cross-graph misuse.
> 
> **Makes TestFlight native crashes readable in Sentry:** Sentry Cocoa (`NativeDiagnostics.swift`) starts from Rust with a production-only DSN, heavy scrubbing, and crash/hang/watchdog capture. Release builds set `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`, verify executable/dSYM UUIDs, upload dSYMs via `sentry-cli`, and fail misconfigured partial Sentry env. The TestFlight workflow now requires Sentry secrets, zips the `.xcarchive`, and uploads versioned IPA/archive artifacts (90-day retention).
> 
> **Desktop/mobile polish:** the file watcher uses a **pruned file-ID cache** (skips `.git`/`.reflect`/prune dirs) to cut watch startup cost; `usePoll` pauses while hidden; back-swipe attaches a gesture-scoped non-passive `touchmove` blocker; editor mount logic is centralized in `whenEditorMounted`. Removes the Plan 19 `spike_mobile` startup probe. Bumps beta to `0.8.0-beta.4`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5459623ef5a239e6abe7846e3b8b9ef59293164. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/desktop/src-tauri/src/db/mod.rs
+++ b/apps/desktop/src-tauri/src/db/mod.rs
@@ -24,7 +24,7 @@ use std::sync::{Mutex, MutexGuard};
 
 use rusqlite::{params, Connection};
 use serde_json::{Map, Value};
-use tauri::State;
+use tauri::{Manager, State};
 
 use crate::background_task::{self, BackgroundTaskState};
 use crate::error::{AppError, AppResult};
@@ -343,35 +343,42 @@ pub fn index_move<R: tauri::Runtime>(
 /// returns the empty scan — that pass is superseded and its writes would be
 /// dropped anyway, so "nothing to do" is the honest answer.
 #[tauri::command]
-pub fn index_reconcile_scan(
+pub async fn index_reconcile_scan<R: tauri::Runtime>(
     generation: u64,
-    graph: State<GraphState>,
-    index: State<IndexState>,
+    app: tauri::AppHandle<R>,
 ) -> AppResult<scan::ReconcileScan> {
-    let started = std::time::Instant::now();
-    let root = crate::fs::current_root(&graph)?;
-    let walk_started = std::time::Instant::now();
-    let files = crate::fs::note_files(&root);
-    let walk_ms = walk_started.elapsed().as_millis() as u64;
-    let state = lock_state(&index)?;
-    if state.generation != generation {
-        return Ok(scan::ReconcileScan::empty());
-    }
-    let conn = state.conn.as_ref().ok_or_else(AppError::no_graph)?;
-    let now_ms = std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .map(|elapsed| elapsed.as_millis() as u64)
-        .unwrap_or(0);
-    let scan = scan::scan_reconcile(conn, &files, now_ms)?;
-    tracing::info!(
-        files = scan.total,
-        candidates = scan.candidates.len(),
-        orphans = scan.orphans.len(),
-        walk_ms,
-        total_ms = started.elapsed().as_millis() as u64,
-        "index_reconcile_scan"
-    );
-    Ok(scan)
+    // Async on purpose: this is one uninterruptible O(graph) walk + table
+    // read, and as a sync command it occupied the iOS main thread — the
+    // post-paint "I can see a note but can't tap it" freeze on every open.
+    crate::blocking::run_blocking(move || {
+        let started = std::time::Instant::now();
+        let graph = app.state::<GraphState>();
+        let root = crate::fs::current_root(&graph)?;
+        let walk_started = std::time::Instant::now();
+        let files = crate::fs::note_files(&root);
+        let walk_ms = walk_started.elapsed().as_millis() as u64;
+        let index = app.state::<IndexState>();
+        let state = lock_state(&index)?;
+        if state.generation != generation {
+            return Ok(scan::ReconcileScan::empty());
+        }
+        let conn = state.conn.as_ref().ok_or_else(AppError::no_graph)?;
+        let now_ms = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map(|elapsed| elapsed.as_millis() as u64)
+            .unwrap_or(0);
+        let scan = scan::scan_reconcile(conn, &files, now_ms)?;
+        tracing::info!(
+            files = scan.total,
+            candidates = scan.candidates.len(),
+            orphans = scan.orphans.len(),
+            walk_ms,
+            elapsed_ms = started.elapsed().as_millis() as u64,
+            "index_reconcile_scan"
+        );
+        Ok(scan)
+    })
+    .await
 }
 
 /// One `index_touch` entry: re-stamp `path`'s stored mtime to `mtime`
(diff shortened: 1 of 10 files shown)


apps/desktop/src-tauri/src/db/mod.rs
```diff
@@ -343,34 +343,41 @@ pub fn index_move<R: tauri::Runtime>(
...
+            files = scan.total,
+            candidates = scan.candidates.len(),
+            orphans = scan.orphans.len(),
+            walk_ms,
+            elapsed_ms = started.elapsed().as_millis() as u64,
+            "index_reconcile_scan"
+        );
+        Ok(scan)
+    })
+    .await
```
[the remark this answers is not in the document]
maccman: Real race — fixed in 41472510. `index_open` now records the root it opened for (rebound with the connection under one lock), the scan lists that pinned root instead of the graph state's current one, and the generation is re-checked after the walk. Regression test covers the exact pre-`index_open` window.

apps/desktop/src-tauri/src/db/mod.rs
```diff
@@ -543,13 +572,22 @@ pub fn embed_remove(
...
     sql: String,
     params: Vec<Value>,
-    index: State<IndexState>,
+    app: tauri::AppHandle<R>,
 ) -> AppResult<Vec<Map<String, Value>>> {
-    let state = lock_state(&index)?;
-    let conn = state.conn.as_ref().ok_or_else(AppError::no_graph)?;
-    query::run_query(conn, &sql, &params)
+    crate::blocking::run_blocking(move || {
+        let index = app.state::<IndexState>();
```
maccman, line 635: `db_query` needs to pin the index session before the first await. Unlike the old synchronous command, this closure can start after a later `index_open` has rebound `IndexState`. All Kysely calls omit a generation, so an in-flight query for graph A can then read graph B and let TanStack Query cache B's rows under graph A's root-scoped key (`staleTime: Infinity`). `openRecent` only stops the index lifecycle; it does not cancel these mounted UI queries. Please snapshot the current index generation before scheduling and reject if it changed in the closure, or plumb the generation through `db_query`.

maccman (commented):
I found two additional issues in the new `db_query` path. The existing unresolved `list_files` graph-snapshot thread also still applies.

apps/desktop/src-tauri/src/db/mod.rs
```diff
@@ -543,13 +572,22 @@ pub fn embed_remove(
...
+        let state = lock_state(&index)?;
```
maccman: Moving the reader off-main while holding this same mutex still lets synchronous writers freeze the main thread. `index_apply`, `index_touch`, `index_clear`, etc. remain sync commands; if one arrives while this query owns `IndexState`, its `lock_state` blocks the iOS main thread until the read finishes. Mounted startup queries can overlap reconcile writes, and normal searches can overlap saves, recreating the exact touch-delivery stall this PR is intended to remove. This needs a reader connection/pool (or another design where sync commands never wait for a background-held lock), not just an off-main reader.

apps/desktop/src-tauri/src/db/mod.rs
maccman: Fixed in 323175d6 via the snapshot option: `db_query` pins the read session's generation before the hop and errors inside the closure if it changed — an erroring superseded query is honest, since its observers unmounted with the switch and nothing retries it against the wrong graph. The pin lives on the new read-side state (see the sibling thread), which `index_open` rebinds generation-and-connection together under one lock.

apps/desktop/src-tauri/src/db/mod.rs
maccman: Agreed — the shared mutex reintroduced the stall through the back door. Fixed in 323175d6: `db_query` now runs on a dedicated **read-only** WAL connection (`ReadInner`) under its own lock, opened by `index_open` right after the writer (writer → reader lock order, nothing locks the reverse way). Sync write commands only ever touch the writer lock, so they no longer wait behind a background read; WAL readers see the last committed state, so a query after an awaited write still reads its result. The read-only open lives in `reflect-index-schema` (`open_index_read_only_at`) next to the writer opener so the CLI/desktop schema contract stays in one place.

apps/desktop/src-tauri/src/fs/mod.rs
```diff
@@ -822,9 +826,19 @@ fn move_to_graph_trash(root: &Path, abs: &Path) -> AppResult<()> {
...
-    Ok(file_catalog(&state, generation)?.notes)
+pub async fn list_files<R: tauri::Runtime>(
+    generation: Option<u64>,
+    app: tauri::AppHandle<R>,
+) -> AppResult<Vec<FileMeta>> {
+    crate::blocking::run_blocking(move || {
+        let state = app.state::<GraphState>();
+        Ok(file_catalog(&state, generation)?.notes)
+    })
+    .await
```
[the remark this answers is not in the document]
maccman: Landed the pin in 195e7256: unpinned calls now capture the current graph generation before the hop, and `file_catalog`'s existing check rejects a superseded call loudly instead of listing the swapped-in root.

PR #1012, opened 2026-08-01
fix: symbolicate native iOS crashes from TestFlight builds

maccman:
## Problem

TestFlight crash feedback is unreadable today. Tester feedback bundles carry a screenshot and device metadata, not a stack, and the failures that matter most on iOS — Rust panics, Swift crashes, main-thread hangs, watchdog/jetsam kills — never reach the JavaScript exception reporter at all, so nothing else records them either.

Even the Apple `.ips` path is broken for Reflect:

- Xcode Organizer symbolicates from the **local** `.xcarchive`, which a CI runner deletes at the end of the run.
- Rust release builds default to `debug = false`, so the archive's dSYM has no line tables for any Rust frame.

## Approach

This is a smaller, non-user-facing slice of #942 — the native-backtrace half only. It adds no UI, no settings entry, no share sheet, no recovery screen, no local journal, and no startup gating; app behavior is unchanged apart from the SDK start.

## Changes

1. **Native iOS crash reporting.** `NativeDiagnostics.swift` starts Sentry Cocoa (pinned `8.58.4` — Sentry 9 requires iOS 15, above Reflect's iOS 14 target) early in Tauri's mobile setup, and `native_diagnostics.rs` gates that on the compile-time production DSN so forks and local builds never initialize it. Enabled: crashes, fully blocking app hangs, watchdog terminations, converted MetricKit diagnostics. Disabled: sessions, breadcrumbs, tracing, profiling, network/file instrumentation, screenshots, view hierarchy, raw MetricKit payloads, PII. Events without a stack are dropped; exception text is redacted, source paths, context lines, frame variables and thread names are stripped, and loaded-image paths are reduced to basenames while keeping the UUID/address data symbolication needs.

2. **Symbols that survive the build.** `release-ios.mjs` builds Rust with `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`, requires the current archive's main dSYM, verifies its Mach-O UUID against the archived executable, and uploads that archive's dSYMs with `sentry-cli --no-sources`. Partial or foreign Sentry configuration fails before the long build rather than after it, and a failed symbol upload blocks publication.

3. **Archive retention.** The TestFlight workflow packages the exact `.xcarchive` and keeps it, plus the build-numbered IPA, for 90 days — the fallback for an Apple `.ips` that arrives weeks later. Packaging runs on `always()`, so the evidence survives a failed run.

4. **Docs.** `docs/privacy.md` gets the native allow-list; `docs/ios-testflight.md` documents symbol retention and the "TestFlight feedback has no stack" / "`.ips` is unsymbolicated" cases.

## Tests

- Release-helper tests cover the Rust debug setting (including that an explicit override wins), the `sentry-cli` args, production DSN validation, partial credentials, and `dwarfdump --uuid` parsing.
- Rust unit tests cover production-DSN validation.

## Verification

- `pnpm check` — passed (one pre-existing unrelated warning)
- `pnpm test --run apps/desktop/scripts/release-ios.test.mjs` — 20 passed
- `cargo test -p reflect-open native_diagnostics --lib` — passed
- `cargo fmt --all -- --check`, `cargo clippy -p reflect-open --lib --all-targets -- -D warnings` — passed
- `cargo check -p reflect-open --lib --target aarch64-apple-ios` — passed
- `swiftc -typecheck -target arm64-apple-ios14.0` against Sentry Cocoa 8.58.4 — passed
- `parseDwarfdumpUuids` run against a real local `.xcarchive`: executable and dSYM both resolve to `BE28A389-…`, and the archive/dSYM paths the helper asserts all exist
- Xcode resolved Sentry Cocoa `8.58.4` (`Package.resolved` committed)

## Risk / Rollout

- No schema or migration impact; nothing here can block note startup.
- `SENTRY_DSN` and `SENTRY_AUTH_TOKEN` were already required by the TestFlight workflow, but the token now also needs **debug-file upload** access to `reflect-open`. Without it, TestFlight publication fails loudly rather than shipping an unsymbolicated build.
- Sentry Cocoa is pinned at `8.58.4` while the app supports iOS 14; moving to Sentry 9 is a separate minimum-iOS decision.
- Compile/link/typecheck verified, but no intentional device crash was generated here. After the first TestFlight build, confirm one symbolicated native event in Sentry with readable Rust frames.

---

> [!NOTE]
> **Medium Risk**
> Touches the TestFlight release gate (Sentry token needs debug-file scope) and adds third-party crash SDK at app startup, but startup is not gated on it and misconfiguration fails the build rather than shipping blind.
> 
> **Overview**
> Makes **TestFlight native failures** (Rust/Swift crashes, hangs, watchdog kills) readable by wiring **scrubbed Sentry Cocoa** into the iOS host and hardening the **release symbol pipeline**.
> 
> **Native reporting:** Adds `NativeDiagnostics.swift` (Sentry **8.58.4**) started from Rust during Tauri mobile setup, gated on the compile-time production `VITE_SENTRY_DSN`. Events are stack-only with aggressive redaction; release/dist align with the WebView SDK.
> 
> **Release / CI:** `release-ios.mjs` sets `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`, verifies executable vs dSYM UUIDs, uploads native dSYMs via `@sentry/cli` (`--no-sources`), and **fails** on partial/wrong Sentry config or failed upload. TestFlight workflow **zips and retains** the `.xcarchive` plus build-numbered IPA artifacts for **90 days**.
> 
> **Docs:** `privacy.md` and `ios-testflight.md` cover native diagnostics, symbol retention, and troubleshooting unsymbolicated `.ips` reports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 094272477a1dd98fe57218017bc813585e885b68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added native iOS crash diagnostics, including watchdog terminations and MetricKit reports.
  - Added automatic symbol validation and Sentry dSYM uploads for iOS releases.
  - Build artifacts now include resolved build numbers, with archives retained for later symbolication.
  - Added safeguards to verify production diagnostics configuration before reporting native crashes.

- **Documentation**
  - Updated TestFlight setup, troubleshooting, privacy, metadata, and diagnostic data documentation.

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -102,9 +102,11 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Resolve build number
+        id: build_number
         run: |
           build_number="$(date -u +%Y%m%d%H%M)"
           echo "BUILD_NUMBER=$build_number" >> "$GITHUB_ENV"
+          echo "value=$build_number" >> "$GITHUB_OUTPUT"
           echo "Using TestFlight CFBundleVersion $build_number"
 
       - name: Build and upload to TestFlight
@@ -124,12 +126,37 @@ jobs:
           fi
           pnpm release:ios "${args[@]}"
 
+      # Xcode Organizer symbolicates a TestFlight crash report from the local
+      # archive, which a CI runner throws away. Keep it so an Apple `.ips` that
+      # arrives weeks later can still be read.
+      - name: Package the Xcode archive
+        if: always()
+        run: |
+          archive="apps/desktop/src-tauri/gen/apple/build/reflect-open_iOS.xcarchive"
+          if [ ! -d "$archive" ]; then
+            echo "::warning::No Xcode archive was produced; skipping symbol packaging"
+            exit 0
+          fi
+          ditto -c -k --sequesterRsrc --keepParent \
+            "$archive" "$RUNNER_TEMP/reflect-ios-xcarchive-$BUILD_NUMBER.zip"
+
       # The exported IPA, kept for post-hoc signing/entitlements inspection —
       # TestFlight builds cannot be downloaded back from App Store Connect.
       - name: Upload IPA artifact
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: reflect-ios-ipa
+          name: reflect-ios-ipa-${{ steps.build_number.outputs.value }}
           path: apps/desktop/src-tauri/gen/apple/build/arm64/*.ipa
           if-no-files-found: ignore
+          retention-days: 90
+
+      - name: Upload Xcode archive artifact
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: reflect-ios-xcarchive-${{ steps.build_number.outputs.value }}
+          path: ${{ runner.temp }}/reflect-ios-xcarchive-${{ steps.build_number.outputs.value }}.zip
+          if-no-files-found: ignore
+          retention-days: 90
+          compression-level: 0
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -60,7 +60,7 @@
     "zod": "4.4.3"
   },
   "devDependencies": {
-    "@sentry/cli": "2.58.6",
+    "@sentry/cli": "3.6.2",
     "@sentry/vite-plugin": "^5.3.0",
     "@sqlite.org/sqlite-wasm": "3.53.0-build1",
     "@tailwindcss/vite": "^4.3.1",
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -60,6 +60,7 @@
     "zod": "4.4.3"
   },
   "devDependencies": {
+    "@sentry/cli": "2.58.6",
     "@sentry/vite-plugin": "^5.3.0",
     "@sqlite.org/sqlite-wasm": "3.53.0-build1",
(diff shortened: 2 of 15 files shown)


maccman:
Thanks — addressed three of the four:

- **`docs/privacy.md` scope (major):** valid, fixed. The native SDK is iOS-only; the summary line now says so instead of implying desktop sends native crash data.
- **`@sentry/cli` version:** valid, bumped `2.58.6` → `3.6.2`. Verified against the 3.6.2 binary that every flag this script passes is unchanged (`--org`, `--project`, `--type dsym`, `--no-sources`, `--wait-for`), and `--no-sources` has identical help text in both. The 3.0 breaking changes are the dropped legacy `SENTRY_API_KEY` auth (we use `SENTRY_AUTH_TOKEN`), named exports in the JS wrapper (we invoke the binary, not the wrapper), and Node ≥ 18 (CI is on 22).
- **Duplicated DSN identity:** fair — the org/project is now asserted in three languages. Added cross-references in `isProductionSentryDsn` and `parse_native_dsn` naming all three sites. I skipped the cross-language test: the two new copies would still have to be edited by hand for a rotation, so a test would only restate the comment at the cost of a build-time coupling between the release script and the Rust crate.

**Skipped — `dwarfdump` try/catch:** this matches the convention in `release-ios.mjs`, which reserves `fail()` for semantic verification failures and lets a broken tool throw. `readIpaInfoPlistRawValue` does exactly the same with `unzip` and `plutil`. The three `existsSync` guards already cover the reachable causes, and either way the release stops with a nonzero exit — only the message quality differs. Wrapping just this one call site would make the file less consistent, not more.

PR #1093, opened 2026-08-10
fix: match partial queries across note titles and content

### LINKED ISSUE #1089: Partial match doesn't work for note content
Partial match works for note title:

Image

Partial match doesn't work for note content:

Image

maccman:
## Problem

Closes https://github.com/team-reflect/reflect-open/issues/1089 

Note content search required whole tokens even though note subjects matched while a query was still being typed. The first fix added body prefixes, but a query split across a partial subject term and a partial body term still failed.

For example, `authent migr` did not find `authentication migration`, and `secur migr` did not find a note titled `Security rollout` whose body contains `migration`.

## Before -> After

| Query shape | Before | After |
| --- | --- | --- |
| Partial body terms | Required complete body tokens | Match from the start of each body word |
| Partial title + partial body | Could miss because title FTS required a whole token | Terms can match independently across title and body word prefixes |
| Mid-word terms | Did not match | Still do not match |
| Title-only presentation | Partial typed range, no unrelated body snippet | Unchanged |

### Before, partial title, 👍

before-title

### Before, partial body, 👎

before-body

### After, partial title, 👍

after-title

### After, partial body, 👍

after-body

## Changes

1. Build every escaped FTS term as a controlled title-prefix or body-prefix expression, preserving explicit `AND` semantics and keeping user-entered FTS operators inert.
2. Merge manual title-recall ranges with FTS ranges so a partial title match highlights only the characters the user typed, including mixed title/body queries.
3. Use body highlight markers to discard marker-free SQLite fragments and keep ordinary title-prefix recall out of BM25 tiebreaking; tokenizer-normalized title matches such as `cafe` → `Café` retain their title-weighted lexical rank.
4. Mirror the behavior in the `reflect` CLI, regenerate the TypeScript/Rust parity fixture, and document cross-column word-prefix matching.

## Tests

- Real SQLite coverage verifies `authent migr`, `secur migr`, and reversed `migr secur` queries, plus the `thentication` mid-token negative case.
- Title highlighting coverage verifies prefix tokens are clipped to the typed range without regressing tokenizer-normalized matches.
- Desktop and CLI integration coverage pins null/empty title-only snippets, score/ranking behavior, pinned title-prefix tiebreaks, and tokenizer-normalized title ranking above body hits.
- TypeScript/Rust query-builder parity remains byte-for-byte covered.

## Verification

- `pnpm check`
- `pnpm test --run packages/core/src/indexing/search-query.test.ts packages/core/src/indexing/title-highlight.test.ts packages/core/src/indexing/filtered-search.test.ts packages/core/src/indexing/pipeline.test.ts packages/core/src/indexing/parity.test.ts apps/desktop/src/dev/dev-index-db.test.ts` — 81 passed
- `cargo fmt --check`
- `cargo test -p reflect-cli search::tests`
- `cargo clippy -p reflect-cli --all-targets -- -D warnings`
- `cargo test -p reflect-cli --test cli search_` — 15 passed
- `cargo test -p reflect-cli --test parity fts_match_building_matches_the_ts_pipeline`

## Risk / Rollout

No schema migration, reindex, or backfill is required; this changes query construction and result presentation only. Recall intentionally expands at word starts, while arbitrary infix matches remain excluded. A minimum prefix length was deliberately not added without production profiling because it would remove legitimate one-character and unsegmented-script searches; that performance policy can be evaluated separately.

---

> [!NOTE]
> **Medium Risk**
> Changes shared search query construction and ranking across core, desktop, and CLI without a reindex, so recall and ordering behavior shifts for all indexed graphs; risk is mitigated by TS/Rust parity tests and broad SQLite/integration coverage.
> 
> **Overview**
> Search now treats each query term as a **word-start prefix** in both title and body FTS columns (`(title : "term"* OR body : "term"*)` with explicit `AND`), so partial typing works across columns—for example `authent migr` can match `authentication` in the body and `secur migr` can match a partial title plus `migration` in the body. User-entered FTS operators stay literal inside quoted terms.
> 
> **Ranking and presentation** align desktop and CLI: title-only hits admitted by manual recall keep **score `0`** and **no snippet**; tokenizer-normalized title matches (e.g. `cafe` → `Café`) still use bm25 above body hits. Snippets are shown only when the body fragment contains highlight markers; marker-free SQLite fragments are dropped.
> 
> **Title highlighting** merges recall with FTS ranges, clips prefix highlights to what the user typed (mixed queries like `secur migr`), and allows partial multi-term recall when not every term matches.
> 
> Parity fixture, CLI docs, and integration tests cover prefix bodies, mixed title/body queries, pinned title-prefix ties, and mid-token negatives (`thentication`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9af6164c7ba71a5d4441aa31719130111e7460c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Search now supports word-prefix matching across note titles and body content.
  * Multi-term queries can match across fields, with highlighted body snippets.
  * Title-only matches retain appropriate ranking and pinned-note ordering.
  * Accent-normalized and Unicode-aware search improves matching across languages.

* **Bug Fixes**
  * Improved handling of punctuation, operators, quoted terms, whitespace, and ignored search terms.
  * Title-only results no longer display misleading snippets or scores.

* **Documentation**
  * Updated CLI search documentation to explain prefix matching, snippets, and result behavior.


packages/core/src/indexing/search-query.ts
```diff
@@ -18,16 +19,22 @@ export function splitSearchTerms(query: string): string[] {
...
 export function buildFtsMatch(query: string): string | null {
   const terms = splitSearchTerms(query)
   if (terms.length === 0) {
     return null
   }
-  return terms.map((term) => `"${term.replaceAll('"', '""')}"`).join(' ')
+  return terms
+    .map((term) => {
+      const literal = `"${term.replaceAll('"', '""')}"`
+      return `(title : ${literal} OR body : ${literal}*)`
```
maccman: **Perf (non-blocking):** `search_fts` is declared without a `prefix=` option, so every `body : "t"*` term is answered by a term-index range scan + doclist merge. The palette runs this on each keystroke, and the very first keystroke produces a 1-char prefix (`body : "a"*`) that can match a large fraction of the body index before bm25/limit apply.

Probably fine at typical graph sizes, but if palette latency ever shows up on large graphs, two cheap options:

- only append `*` when `term.length >= 2` (or 3) — single chars fall back to exact-token matching, which is close to useless for recall anyway; or
- add `prefix='2 3'` to the FTS table in a later migration (that one does need a reindex, which this PR deliberately avoids).

Whatever you pick, remember the Rust twin in `apps/cli/src/search.rs` and the parity fixture must move together.

apps/desktop/src/dev/dev-index-db.test.ts
```diff
@@ -263,6 +263,31 @@ describe('createDevIndexDb', () => {
     expect(hits[2]!.highlightedTitle).toBe('Garage')
   })
 
+  it('prefix-matches partial terms in note bodies', async () => {
```
maccman, line 266: **Observation (non-blocking):** there's a residual gap this test suite doesn't pin down: a query mixing a partial *title* word with a partial *body* word still misses. With this fixture, `secur migr` returns nothing — `title : "secur"` needs the whole token, `body : "secur"*` finds no body word starting with `secur`, so the FTS `AND` fails; and the title-recall path requires *all* terms to match the title, which `migr` doesn't.

That's consistent with the PR's stated scope (title semantics unchanged), and prefix-matching the title inside FTS would change `highlight()` ranges, so the tradeoff makes sense. But it might be worth a negative test asserting this behavior (like the `thentication` guard) so a future change to it is deliberate rather than accidental — or noting it as a known follow-up.

maccman (commented):
Reviewed the query-construction change plus the CLI mirror, parity fixture, schema (`fts5(path UNINDEXED, title, body)`), and both consumers (`filtered-search.ts`, `search_index` in the CLI).

**Correctness** — looks right. For each term the new expression is a strict superset of the old recall: the old whole-token match in either column is covered by `title : "t"` (exact) or `body : "t"*` (prefix includes exact), so nothing that matched before stops matching. The switch from implicit adjacency to explicit `AND` is required once terms become parenthesized subexpressions, and semantics are identical. Quoting/escaping is unchanged, and the trailing `*` sits outside the quoted literal, so user-typed operators stay inert. Since `path` is `UNINDEXED`, column-scoping to `title`/`body` loses nothing. Highlight/snippet column indexes (1 = title, 2 = body) are untouched.

**Tests/docs** — good coverage: real-SQLite regression with a mid-token negative guard, CLI integration test, regenerated parity fixture, and `docs/cli.md` now describes the behavior accurately.

Two non-blocking observations, left inline:

1. Short-prefix body queries (`body : "a"*` on the first keystroke) with no `prefix=` index on `search_fts` — worth keeping an eye on for palette latency on large graphs.
2. A residual recall asymmetry: a query mixing a *partial title* word with a *partial body* word (e.g. `secur migr`) still returns nothing, because the FTS title branch is exact-token and the title-recall path requires all terms in the title. Fine if intentional — flagging so it's a known follow-up rather than a surprise.

packages/core/src/indexing/search-query.ts
maccman: Valid concern, but I’m leaving the prefix-length policy out of this PR. A global 2–3 character cutoff would remove legitimate one-character and CJK/unsegmented-script recall, while an FTS `prefix=` migration would not help the broad one-character case. The current search surfaces do accept one-character queries, so this is worth profiling separately against production-sized graphs before choosing a policy. No code change here.

apps/desktop/src/dev/dev-index-db.test.ts
maccman: Fixed in 209866cb. Each term now supports title or body word-prefix matching, so `secur migr` and the reversed `migr secur` both find a `Security…` title with `migration` in the body. I also preserved the prior presentation semantics: only the typed title prefix is highlighted, title-only hits do not show an unrelated body fragment, and title-only ranking uses the existing deterministic tiebreakers. Added real-SQLite and CLI coverage for the mixed query.

packages/core/src/indexing/filtered-search.ts
```diff
@@ -307,16 +307,23 @@ export async function searchWithFilters(
     ])
     .where(sql<boolean>`("lexical"."path" is not null or ${titleMatch.containsAllTerms})`)
     .orderBy(titleMatch.rank)
-    .orderBy(sql`coalesce("lexical"."rank", 0)`)
+    // A title-only lexical prefix has no body marker. Keep its historical
+    // title-recall rank (`0`) rather than letting lexical bm25 reorder it.
+    .orderBy(
+      sql`case when instr(coalesce("lexical"."snippet", ''), ${HIGHLIGHT_START}) > 0 then coalesce("lexical"."rank", 0) else 0 end`,
+    )
```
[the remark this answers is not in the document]
maccman: Fixed in 9af6164c. The BM25 neutralization now applies only when manual title recall already admits the hit. Tokenizer-only title matches such as `cafe` → `Café` retain their title-weighted lexical rank (and CLI score), while ordinary title-prefix recall keeps the deterministic zero-score tiebreak behavior. Added real-SQLite and CLI regressions that place the normalized title hit above a body hit.

ocavue (changes requested):
Blocked by https://github.com/team-reflect/reflect-open/pull/1114

ocavue (approved):
LGTM

PR #1163, opened 2026-08-22
feat: add on-device page summaries to Chrome captures

maccman:
## Problem

Reflect's Chrome extension can capture a page's complete readable text, but it cannot turn that text into a concise local summary. Waiting for a summary before saving would also make link capture fragile: closing the popup or waiting for Chrome's local model download could lose the user's primary action.

The keyboard shortcut currently bypasses the popup, which leaves no document context or fresh Enter/click gesture for Chrome's Summarizer API. The API is unavailable in the MV3 service worker.

## Before -> After

| Before | After |
| --- | --- |
| Capture the link or full page text | Independently capture full text, an on-device summary, or both |
| Wait for optional page-text extraction before the only save | Queue the raw link first, then refresh the same capture with the summary |
| Command-Shift-K saves immediately with stored defaults | Command-Shift-K opens the focused popup; Enter saves with the visible options |
| Captured notes have Note, Selection, Page Text, and Screenshot sections | Browser-generated summaries have a dedicated Summary section |

## Changes

1. **On-device summary flow**
   - Feature-detects Chrome's Summarizer API and model availability without raising the extension's minimum Chrome version.
   - Starts model creation from the form submission gesture, reports model-download/generation progress, observes the model input quota, and releases or aborts the session with the popup lifecycle.
   - Persists an independent Capture page summary preference. Summary captures queue the raw link first and send a same-timestamp recapture afterward so existing URL/selection deduplication updates the original note.
   - Preserves the existing page-text option; if summary generation fails, requested page text still lands and the raw link is never lost.

2. **Capture contract and note shape**
   - Adds an optional `summary` field to the additive version-1 capture envelope, extension wire builder, Rust native host, and shared parity fixtures.
   - Renders the generated Markdown under `## Summary`, after the user's note and before selection/page text.
   - Treats a drain-managed summary refresh during desktop enrichment as a pending managed update rather than a user edit, so metadata/BYOK enrichment can retry without clobbering the summary.

3. **Shortcut and disclosures**
   - Uses `action.openPopup()` for the keyboard command, with the old instant capture as a pre-Chrome-127/rejected-popup fallback.
   - Updates extension/store copy, the link-capture design record, and privacy disclosures to distinguish local Chrome AI from provider-backed AI and explain the one-time model download.

## Tests

- Summarizer feature detection, availability, configuration, download progress, quota trimming, empty output, cancellation, and cleanup.
- Two-phase ordering, stable capture time, independent text/summary options, failure fallback, and rejected raw links.
- Shortcut popup opening and legacy fallback.
- Extension wire shaping, preferences, shared TypeScript/Rust envelope parity, summary note rendering, same-day deduplication, and summary/enrichment races.

## Verification

- `pnpm check`
- `pnpm test --run apps/extension/lib/page-summary.test.ts apps/extension/entrypoints/popup/capture-flow.test.ts apps/extension/lib/commands.test.ts apps/extension/lib/popup-preferences.test.ts apps/extension/lib/capture-message.test.ts packages/core/src/actions/capture-envelope.test.ts packages/core/src/actions/capture-envelope.parity.test.ts packages/core/src/actions/capture-drain.test.ts packages/core/src/actions/capture-enrichment.test.ts` — 161 passed
- `cargo test -p reflect-capture-host` — 23 passed
- `pnpm --filter @reflect/extension build`

## Risk / Rollout

- This is an additive envelope change with no data migration. The extension and desktop/native host should ship together; older hosts tolerate but discard the unknown summary field, while the already-queued raw link remains safe.
- Chrome may download its built-in model on first use. Captured page text stays on-device and is not sent to Reflect or an AI provider.
- Browsers without usable built-in summarization see a disabled option and retain all existing capture behavior.

## Summary by CodeRabbit

* **New Features**
  * Added optional on-device page summaries during browser capture, with selectable page text, summaries, or both.
  * Added progress indicators for model downloads and summary generation.
  * Added a saved preference for enabling summaries.
  * Summaries are included in notes and preserved during queued processing.
  * Keyboard capture opens capture options, with direct-save fallback when unavailable.

* **Bug Fixes**
  * Improved handling of capture failures, cancellations, retries, and concurrent updates.

* **Documentation**
  * Clarified local processing, privacy protections, permissions, and summary behavior.

### DIFF
--- a/apps/extension/README.md
+++ b/apps/extension/README.md
@@ -22,8 +22,9 @@ raw note + daily `## [[Links]]` entry now (resolving or creating that category
 note), meta-scrape + BYOK AI title + description async. Optional summaries run
 inside Chrome with its built-in local model; the extension stores no keys and
 sends no page content to an AI provider. Chrome may download the model before its
-first use. The extension's only honest status is **queued** — it cannot observe
-the desktop drain.
+first use. After local handoff, the desktop app may separately perform BYOK AI
+enrichment under Reflect's desktop privacy rules. The extension's only honest
+status is **queued** — it cannot observe the desktop drain.
 
 ## Develop
 
@@ -131,16 +132,18 @@ in the [Developer Dashboard](https://chrome.google.com/webstore/devconsole).
 >
 > A capture includes the page's URL and title, your current text selection, and a
 > screenshot of the visible tab. Optionally, tick "Capture page text" to include the
-> page's readable text, or "Capture page summary" to summarize that text with Chrome's
-> built-in on-device AI. The link is queued first, then its summary is added shortly
+> page's readable text and/or "Capture page summary" to summarize that text with
+> Chrome's built-in on-device AI. The two options are independent and can be selected
+> together. The link is queued first, then its text and summary are added shortly
 > afterward.
 >
 > Captures are handed to the **installed Reflect desktop app** over a local connection
 > on your own machine — there is no Reflect account and no Reflect server in the path.
 > Capturing works even when the app is closed: the link is held and saved automatically
 > the next time Reflect runs. The extension stores no API keys and sends no captured
-> content to an AI provider. Chrome may download its local summarization model before
-> the first summary.
+> content to an AI provider; Chrome may download its local summarization model before
+> the first summary. After local capture, the desktop app may perform separately
+> configured BYOK AI enrichment under Reflect's privacy controls.
 >
 > Requires the Reflect desktop app: https://github.com/team-reflect/reflect-open
 
@@ -177,8 +180,10 @@ Each is reviewed individually; every permission below is exercised by the code:
   screenshot, and — only when opted in — page text and/or its on-device summary). Collected
   **only on an explicit user action**, never in the background.
 - **Where it goes:** to the user's own machine (the local Reflect desktop app). It is
-  **not** sent to Reflect or an AI provider; Chrome's built-in model processes summary
-  input locally.
+  **not** sent to Reflect, and the extension sends nothing to an AI provider; Chrome's
+  built-in model processes summary input locally. Once captured, the desktop app may
+  separately send eligible content to the user's configured BYOK provider for optional
+  enrichment, subject to Reflect's privacy controls.
 - The three required certifications are all true and can be affirmed:
   1. Data is **not** sold to third parties.
   2. Data is **not** used or transferred for purposes unrelated to the single purpose.
--- a/apps/extension/README.md
+++ b/apps/extension/README.md
@@ -1,9 +1,9 @@
 # Reflect Capture (Chrome extension)
 
-Save the page you're reading into Reflect: ⌘⇧K saves immediately with default
-settings, including the stored page-text preference, while the toolbar button
-opens the capture popup for an optional note. Captures include the page URL,
-title, selection, screenshot, and optional page text when Chrome allows them,
+Save the page you're reading into Reflect: the toolbar button or ⌘⇧K opens the
+capture popup, and Enter saves. Captures include the page URL, title, selection,
+screenshot, and optional page text. On supported devices, Chrome's on-device
+Summarizer can also add a page summary shortly after the link is queued. Captures
 then hand off to the **installed desktop app** through a local native-messaging
 host. No Reflect-hosted services are involved, and capture works even while the
 app is closed: the host spools into the graph's capture inbox
@@ -19,9 +19,11 @@ popup → `chrome.storage` queue → background `sendNativeMessage` →
 `reflect-capture-host` (Tauri sidecar, registered by the desktop app on every
 launch) → capture inbox → desktop drain (`@reflect/core` `actions/capture`):
 raw note + daily `## [[Links]]` entry now (resolving or creating that category
-note), meta-scrape + BYOK AI title + description async. The extension stores no
-keys and makes no AI or network calls; its only honest status is **queued** — it
-cannot observe the desktop drain.
+note), meta-scrape + BYOK AI title + description async. Optional summaries run
+inside Chrome with its built-in local model; the extension stores no keys and
+sends no page content to an AI provider. Chrome may download the model before its
+first use. The extension's only honest status is **queued** — it cannot observe
+the desktop drain.
 
 ## Develop
 
@@ -124,18 +126,21 @@ in the [Developer Dashboard](https://chrome.google.com/webstore/devconsole).
 
 **Detailed description:**
 
-> Reflect Capture saves the page you're reading into Reflect with one click or a
-> keyboard shortcut (⌘⇧K / Ctrl+Shift+K).
+> Reflect Capture saves the page you're reading into Reflect from the toolbar or
+> keyboard shortcut (⌘⇧K / Ctrl+Shift+K). Press Enter in the popup to save.
 >
 > A capture includes the page's URL and title, your current text selection, and a
 > screenshot of the visible tab. Optionally, tick "Capture page text" to include the
-> page's readable text as well.
+> page's readable text, or "Capture page summary" to summarize that text with Chrome's
+> built-in on-device AI. The link is queued first, then its summary is added shortly
+> afterward.
 >
 > Captures are handed to the **installed Reflect desktop app** over a local connection
 > on your own machine — there is no Reflect account and no Reflect server in the path.
 > Capturing works even when the app is closed: the link is held and saved automatically
-> the next time Reflect runs. The extension stores no API keys and makes no AI or
-> network calls of its own.
+> the next time Reflect runs. The extension stores no API keys and sends no captured
+> content to an AI provider. Chrome may download its local summarization model before
+> the first summary.
 >
 > Requires the Reflect desktop app: https://github.com/team-reflect/reflect-open
 
@@ -160,7 +165,7 @@ Each is reviewed individually; every permission below is exercised by the code:
 | Permission | Why it's needed |
 | --- | --- |
 | `activeTab` | Read the URL/title and grab a screenshot + selection of the tab you capture — only at the moment you click the button or press the shortcut. Avoids any broad host permission. |
-| `scripting` | Run a one-line script in the active tab to read the current selection and, when opted in, extract the page's readable text. |
+| `scripting` | Run a script in the active tab to read the current selection and, when opted in, extract readable text for full-text capture or on-device summarization. |
 | `nativeMessaging` | The only output: hand each capture to the local `reflect-capture-host` the desktop app registers. No network is used. |
 | `storage` | Queue captures locally so a capture survives the app being closed and retries until it spools. |
 | `unlimitedStorage` | Queued captures embed a screenshot data URL, which can exceed the default storage quota while waiting for the app. |
@@ -169,10 +174,11 @@ Each is reviewed individually; every permission below is exercised by the code:
 ### Data-handling disclosures (Privacy practices tab)
 
 - **Data collected:** *Website content* (the captured page's URL, title, selection,
-  screenshot, and — only when opted in — page text). Collected **only on an explicit
-  user action**, never in the background.
+  screenshot, and — only when opted in — page text and/or its on-device summary). Collected
+  **only on an explicit user action**, never in the background.
 - **Where it goes:** to the user's own machine (the local Reflect desktop app). It is
-  **not** sent to Reflect or any third party.
+  **not** sent to Reflect or an AI provider; Chrome's built-in model processes summary
+  input locally.
 - The three required certifications are all true and can be affirmed:
   1. Data is **not** sold to third parties.
   2. Data is **not** used or transferred for purposes unrelated to the single purpose.
--- a/apps/extension/entrypoints/background.ts
+++ b/apps/extension/entrypoints/background.ts
@@ -54,6 +54,7 @@ export default defineBackground(() => {
 
   browser.commands.onCommand.addListener((command, tab) => {
     if (command === SAVE_CURRENT_PAGE_COMMAND) {
+      void flushQueue()
       const openPopup =
         typeof browser.action.openPopup === 'function'
           ? () => browser.action.openPopup()
--- a/apps/extension/entrypoints/background.ts
+++ b/apps/extension/entrypoints/background.ts
@@ -1,6 +1,6 @@
 import { browser } from 'wxt/browser'
 import { defineBackground } from '#imports'
-import { SAVE_CURRENT_PAGE_COMMAND } from '@/lib/commands'
+import { openCapturePopupOrFallback, SAVE_CURRENT_PAGE_COMMAND } from '@/lib/commands'
 import { flushQueue } from '@/lib/flush'
 import { isFlushRequest } from '@/lib/messages'
 import { readIncludePageTextPreference } from '@/lib/popup-preferences'
@@ -9,11 +9,10 @@ import { snapshotTab } from '@/lib/snapshot-active-tab'
 import { tryExtractPageText } from './popup/extract-page-text'
 
 /**
- * The MV3 service worker owns retries and the shortcut fast path. Every
- * capture is persisted before a flush starts, so nothing depends on this
- * worker's (or the popup's) lifetime. Retries ride four triggers: every flush
- * ping, the keyboard shortcut, browser startup, and a coarse alarm for the
- * "Reflect installed an hour later" case.
+ * The MV3 service worker owns retries and opens the popup for the shortcut.
+ * Every capture is persisted before a flush starts, so nothing depends on
+ * this worker's (or the popup's) lifetime. Retries ride flush pings, browser
+ * startup, and a coarse alarm for the "Reflect installed an hour later" case.
  */
 
 const RETRY_ALARM = 'capture-retry'
@@ -55,9 +54,15 @@ export default defineBackground(() => {
 
   browser.commands.onCommand.addListener((command, tab) => {
     if (command === SAVE_CURRENT_PAGE_COMMAND) {
-      void saveTabWithDefaults(tab).catch((cause: unknown) => {
-        console.error('shortcut capture failed:', cause)
-      })
+      const openPopup =
+        typeof browser.action.openPopup === 'function'
+          ? () => browser.action.openPopup()
+          : undefined
+      void openCapturePopupOrFallback(openPopup, () => saveTabWithDefaults(tab)).catch(
+        (cause: unknown) => {
+          console.error('shortcut capture failed:', cause)
+        },
+      )
     }
   })
 
--- a/apps/extension/entrypoints/popup/capture-flow.ts
+++ b/apps/extension/entrypoints/popup/capture-flow.ts
@@ -118,6 +118,9 @@ export async function runCaptureFlow(
     const pageTextOutcome = await dependencies.saveCapture(
       captureInput(input, dependencies, capturedAt, { contentText }),
     )
+    if (pageTextOutcome.fate === 'rejected') {
+      return { kind: 'update-rejected', linkOutcome }
+    }
     return { kind: 'summary-failed', cause, linkOutcome, pageTextOutcome }
   }
 
--- a/apps/extension/entrypoints/popup/capture-flow.ts
+++ b/apps/extension/entrypoints/popup/capture-flow.ts
@@ -1,0 +1,134 @@
+import type { BuildWireMessageInput, CapturedPage } from '@/lib/capture-message'
+import type { PageSummaryTask } from '@/lib/page-summary'
+import type { SaveOutcome } from '@/lib/save-capture'
+
+export type CaptureFlowPhase =
+  | 'extracting-page'
+  | 'saving-link'
+  | 'generating-summary'
+  | 'saving-update'
+
+export interface CaptureFlowInput {
+  page: CapturedPage
+  tabId: number
+  note: string
+  includePageText: boolean
+  includePageSummary: boolean
+}
+
+export interface CaptureFlowCallbacks {
+  onPhase: (phase: CaptureFlowPhase) => void
+  onDownloadProgress: (progress: number) => void
+}
+
+export interface CaptureFlowDependencies {
+  now: () => Date
+  randomId: () => string
+  extractPageText: (tabId: number, expectedUrl: string) => Promise<string | undefined>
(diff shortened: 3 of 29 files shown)


apps/extension/README.md
```diff
@@ -124,18 +126,21 @@ in the [Developer Dashboard](https://chrome.google.com/webstore/devconsole).
...
+> afterward.
 >
 > Captures are handed to the **installed Reflect desktop app** over a local connection
 > on your own machine — there is no Reflect account and no Reflect server in the path.
 > Capturing works even when the app is closed: the link is held and saved automatically
-> the next time Reflect runs. The extension stores no API keys and makes no AI or
-> network calls of its own.
+> the next time Reflect runs. The extension stores no API keys and sends no captured
+> content to an AI provider. Chrome may download its local summarization model before
+> the first summary.
```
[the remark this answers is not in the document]
maccman: Fixed in 22614ebf: extension/on-device processing is now clearly separated from optional later desktop BYOK enrichment across the README, store disclosure, Plan 11, and privacy table.

apps/extension/entrypoints/popup/capture-flow.ts
```diff
@@ -0,0 +1,134 @@
...
+    return { kind: 'summary-failed', cause, linkOutcome, pageTextOutcome }
+  }
+
+  callbacks.onPhase('saving-update')
+  const updateOutcome = await dependencies.saveCapture(
+    captureInput(input, dependencies, capturedAt, {
+      contentText: input.includePageText ? contentText : undefined,
+      summary,
+    }),
+  )
```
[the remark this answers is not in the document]
maccman: Fixed in 22614ebf: same-day dedup now carries forward managed summary, page text, and description fields from the existing note, with a same-millisecond summary-first regression test.

apps/extension/README.md
```diff
@@ -124,18 +126,21 @@ in the [Developer Dashboard](https://chrome.google.com/webstore/devconsole).
...
-> keyboard shortcut (⌘⇧K / Ctrl+Shift+K).
+> Reflect Capture saves the page you're reading into Reflect from the toolbar or
+> keyboard shortcut (⌘⇧K / Ctrl+Shift+K). Press Enter in the popup to save.
 >
 > A capture includes the page's URL and title, your current text selection, and a
 > screenshot of the visible tab. Optionally, tick "Capture page text" to include the
-> page's readable text as well.
+> page's readable text, or "Capture page summary" to summarize that text with Chrome's
+> built-in on-device AI. The link is queued first, then its summary is added shortly
+> afterward.
```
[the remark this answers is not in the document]
maccman: Fixed in 22614ebf: the store copy and Plan 11 now state that page text and summary are independent options and can be selected together.

apps/extension/entrypoints/popup/capture-flow.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    callbacks.onPhase('saving-update')
+    const pageTextOutcome = await dependencies.saveCapture(
+      captureInput(input, dependencies, capturedAt, { contentText }),
+    )
+    return { kind: 'summary-failed', cause, linkOutcome, pageTextOutcome }
+  }
+
+  callbacks.onPhase('saving-update')
+  const updateOutcome = await dependencies.saveCapture(
```
maccman, line 129: **[correctness] Phase-2 summary recapture clobbers user edits and a `private: true` flag added between phases**

This update save lands as a dedup refresh in `capture-drain.ts` (`writeNote(identity.notePath, await captureNoteSource(envelope, ...))` at ~line 235), which rebuilds the entire note body **and frontmatter** from the envelope. The drain's privacy check reads only the *daily* note (`notePrivate(dailySource)`), never the capture note's own frontmatter.

Scenario: with the app open, phase 1 drains within seconds and creates the note. During a multi-minute model download the user edits the note — or marks it `private: true`. Phase 2 then drains, dedup-matches, and the rewrite erases the edits and the private flag; the refreshed note is stamped `captureStatus: pending` with a self-consistent hash, so enrichment proceeds and sends the content to the BYOK AI provider despite the explicit opt-out.

The wholesale-rewrite path is pre-existing, but it used to fire only on a deliberate manual re-capture; this PR makes it the routine second half of every summary capture, with a window that spans page extraction plus a model download. Related: a summary-enabled re-capture of an already-captured same-day URL strips the existing note's `## Page text` in phase 1 and never restores it if `summarize()` fails with page text unchecked.

Suggested direction: make the refresh a section-level merge (upsert `## Summary` into the existing note) or at minimum skip the refresh when the existing note's frontmatter is private or its body hash no longer matches its own `captureHash` (i.e. user-edited).

apps/extension/entrypoints/popup/capture-flow.ts
```diff
@@ -0,0 +1,134 @@
...
+      : undefined
+    callbacks.onPhase('saving-link')
+    const outcome = await dependencies.saveCapture(
+      captureInput(input, dependencies, capturedAt, { contentText }),
+    )
+    return outcome.fate === 'rejected' ? { kind: 'link-rejected' } : { kind: 'saved', outcome }
+  }
+
+  callbacks.onPhase('saving-link')
+  const linkOutcome = await dependencies.saveCapture(captureInput(input, dependencies, capturedAt))
```
maccman, line 93: **[correctness] Closing the popup mid-flow silently loses explicitly-requested page text**

With summary enabled, this phase-1 save deliberately omits `contentText` — the only saves that carry it are the phase-2 calls after `summarize()` resolves. Chrome destroys action popups on any focus loss, and phase 2 can be minutes away during a model download; the durable queue only ever held phase 1, and nothing retries the update.

Scenario: user ticks both "Capture page text" and "Capture page summary", presses Enter, then clicks back into the page while the popup shows "Downloading Chrome's local model… 12%". The popup unloads, the unmount effect cancels the summary task, and the phase-2 save never happens. The note permanently lacks the page text the user explicitly requested — silently, with no error or retry anywhere. (Pre-PR, and on the current summary-off path, page text is extracted *before* the single durable save.)

Consider including `contentText` in the phase-1 envelope when `includePageText` is on (extraction already has to finish before `summarize()` anyway), so only the summary itself rides the fragile second phase.

apps/extension/entrypoints/popup/capture-flow.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Capture with popup-selected options. Summary captures are deliberately
+ * two-phase: the raw link is durable before page extraction or local model
+ * work begins, then a same-day recapture refreshes that note in place.
+ */
+export async function runCaptureFlow(
+  input: CaptureFlowInput,
+  callbacks: CaptureFlowCallbacks,
+  dependencies: CaptureFlowDependencies,
+): Promise<CaptureFlowResult> {
+  const capturedAt = dependencies.now()
```
maccman, line 72: **[correctness] Spool-order tie can drain the summary before the raw link, wiping the summary**

Both phases share this `capturedAt`, but the drain orders spools by `(modifiedMs, path.localeCompare)` (`capture-drain.ts` ~line 164) — and paths are random UUID filenames. Nothing encodes phase order: no sequence number, and `capturedAt` isn't used for ordering.

Scenario: Reflect isn't running, so both envelopes are held in `chrome.storage`. The next flush replays them back-to-back; if both spool writes land within the same millisecond (fast machine, coarse-mtime filesystem, or the `modified_ms` `unwrap_or(0)` fallback in `capture.rs`), the UUID tiebreak gives ~50% odds the summary envelope drains *first*. The raw-link envelope then drains second as a dedup refresh that rebuilds the note without `## Summary` — silently lost after the popup reported success.

A monotonic sequence in the spool filename, or a drain tiebreak that prefers the envelope carrying more content, would close this.

apps/extension/lib/commands.ts
```diff
@@ -1,2 +1,25 @@
...
+): Promise<OpenCaptureCommandOutcome> {
+  if (openPopup) {
+    try {
+      await openPopup()
+      return 'popup'
+    } catch {
+      // A browser that exposes but rejects openPopup still gets a capture.
+    }
+  }
+  await fallbackCapture()
```
maccman, line 20: **[correctness] Any `openPopup` rejection silently degrades to an instant capture**

`action.openPopup()` can reject transiently on Chrome ≥127 too — unfocused window, a popup already showing — not just on pre-127 browsers as the comment implies. The bare `catch {}` treats every rejection the same: a silent background capture that (necessarily, since the Summarizer API isn't available in the service worker) ignores the summary preference, with no popup, badge, or notification that the user's intent was downgraded.

Worst case: the popup is already open mid-composition, ⌘⇧K rejects, and the fallback queues a second same-URL capture with a fresh id that will dedup-fight the one the user is about to save.

Consider distinguishing "API missing" (fall back) from "API present but rejected" (probably a popup is already open — do nothing or focus it), and surfacing the degrade (e.g. a badge) when the fallback fires.

apps/native-host/src/envelope.rs
```diff
@@ -22,6 +22,8 @@ pub struct Envelope {
     pub selection: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub content_text: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub summary: Option<String>,
```
maccman, line 26: **[correctness] Version skew makes summaries a silent no-op — old hosts strip the field and still ack `queued`**

An already-installed older `reflect-capture-host` tolerates the unknown `summary` field on parse, but `spool_capture` re-serializes the parsed struct (`serde_json::to_vec(&capture.envelope)`), so the field is dropped from the spooled JSON — and the host still acks `queued`, so the extension shows the full success flow.

The skew is the common direction: Chrome auto-updates the extension from the Web Store while the host only updates with the desktop app. Until the user updates the app, every summary capture "succeeds" in the popup but no note ever gets a `## Summary`, and neither side can detect it (`version` stays 1).

The PR description acknowledges the discard, but a detectable signal would help — e.g. a host-version handshake before offering the summary checkbox, or bumping the envelope version so an old host rejects phase 2 (the raw link from phase 1 is already safe).

apps/extension/entrypoints/popup/capture-flow.ts
```diff
@@ -0,0 +1,134 @@
...
+    summary = await summaryTask.summarize(contentText)
+  } catch (cause) {
+    if (!input.includePageText) {
+      return { kind: 'summary-failed', cause, linkOutcome }
+    }
+    callbacks.onPhase('saving-update')
+    const pageTextOutcome = await dependencies.saveCapture(
+      captureInput(input, dependencies, capturedAt, { contentText }),
+    )
+    return { kind: 'summary-failed', cause, linkOutcome, pageTextOutcome }
```
maccman, line 125: **[correctness] A rejected page-text update is reported as a summary-only failure**

Unlike the main path (which maps a rejected phase-2 outcome to `kind: 'update-rejected'`), this fallback returns `'summary-failed'` without checking `pageTextOutcome.fate === 'rejected'`. A rejected envelope is removed from the queue permanently, so the popup's "Link captured, but the summary could not be generated: …" implies the page text landed when it was actually dropped.

Checking `pageTextOutcome.fate` here and routing to the `'update-rejected'` messaging would keep the two paths honest.

apps/extension/lib/page-summary.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+function truncateAtTextBoundary(text: string, maximumLength: number): string {
+  if (text.length <= maximumLength) {
+    return text
+  }
+  const prefix = text.slice(0, Math.max(1, maximumLength))
+  const paragraphAt = prefix.lastIndexOf('\n\n')
+  const wordAt = prefix.lastIndexOf(' ')
+  const boundary = paragraphAt >= prefix.length / 2 ? paragraphAt : wordAt
```
maccman: **[correctness] Word-boundary trim has no floor — over-quota input can collapse to one word**

The paragraph boundary gets a `>= prefix.length / 2` floor, but `wordAt` has none beyond `boundary > 0`. Verified by executing the function: `'A ' + 'x'.repeat(100000)` with `maximumLength` 5000 → `paragraphAt = -1`, `wordAt = 1` → returns `"A"`. That one-character candidate trivially passes `fitToInputQuota`'s measure loop, so Chrome summarizes a one-word input and the note gets a garbage `## Summary`.

Trigger: any page whose extracted text has an early space followed by a long unbroken run — base64/minified blobs, or unspaced CJK preceded by a short Latin word. Applying the same `>= prefix.length / 2` floor to `wordAt` (falling back to a hard cut) fixes it.

apps/extension/entrypoints/background.ts
```diff
@@ -55,9 +54,15 @@ export default defineBackground(() => {
...
   browser.commands.onCommand.addListener((command, tab) => {
     if (command === SAVE_CURRENT_PAGE_COMMAND) {
-      void saveTabWithDefaults(tab).catch((cause: unknown) => {
-        console.error('shortcut capture failed:', cause)
-      })
+      const openPopup =
+        typeof browser.action.openPopup === 'function'
+          ? () => browser.action.openPopup()
+          : undefined
+      void openCapturePopupOrFallback(openPopup, () => saveTabWithDefaults(tab)).catch(
```
maccman, line 62: **[correctness] ⌘⇧K no longer flushes held captures**

Pre-PR, the shortcut path ran `saveCapture(..., flushQueue)`, so every press drained the held queue — the old header comment listed it among the four retry triggers. Post-PR, the successful `openPopup()` path performs no flush: the popup's only queue interaction on open is a read-only `readQueue()` for the held-count line, and a flush happens only if the user actually submits.

Scenario: held captures exist, the desktop app becomes reachable, the user presses ⌘⇧K and dismisses the popup with Esc — "N earlier captures waiting for Reflect" now lingers until the 15-minute `RETRY_ALARM`. Cheap fix: fire a flush ping when the popup opens (or after `openPopup()` resolves).

apps/extension/lib/page-summary.ts
```diff
@@ -0,0 +1,169 @@
...
+  type: 'key-points',
+  format: 'markdown',
+  length: 'medium',
+  preference: 'auto',
+} as const satisfies SummarizerCreateCoreOptions
+
+const QUOTA_HEADROOM = 0.9
+const MAX_QUOTA_TRIM_ATTEMPTS = 4
+
+const summaryOutputSchema = z.string().trim().min(1, 'Chrome returned an empty summary')
```
maccman, line 13: **[correctness] Empty-summary failure surfaces a raw ZodError JSON blob in the popup**

When the model returns whitespace-only output, `summaryOutputSchema.parse` throws a ZodError whose `.message` (zod 4) is the pretty-printed JSON issues array. It propagates uncaught into the flow's `catch`, and `app.tsx` renders `errorMessage(result.cause)` — so the user sees `Link captured, but the summary could not be generated: [ { "origin": "string", "code": "too_small", … } ]` instead of "Chrome returned an empty summary". (Verified against the installed zod 4.4.3; the test only passes because `toThrow` substring-matches inside the blob.)

Either throw a plain `Error` for the empty case or unwrap the first issue's message before display.

packages/core/src/actions/capture-enrichment.ts
```diff
@@ -249,14 +254,16 @@ export async function reconcileCaptureEnrichment(
...
-    if (
-      snapshot.isPrivate ||
-      notePrivate(dailySource) ||
-      bodyHash !== (expectedHash ?? snapshot.meta.captureHash)
-    ) {
+    if (snapshot.isPrivate || notePrivate(dailySource) || bodyHash !== snapshot.meta.captureHash) {
       await markSkipped(snapshot.source, identity)
       return null
     }
+    if (expectedHash !== undefined && snapshot.meta.captureHash !== expectedHash) {
```
maccman, line 261: **[efficiency] The leave-pending carve-out discards a finished BYOK AI call**

This check runs *after* `generateEnrichment` completes, so a summary refresh landing during the provider call throws away the finished result (`return null` → `continue`), and the immediately-scheduled follow-up pass re-runs `scrapePageMeta`, the preview fetch, and a second `describePage` — doubled token spend on the user's key. With the app open this overlap is the common case: the raw link drains instantly, enrichment starts, and the summary lands seconds later.

Since the title/description derive from the page (not the note body), the finished enrichment could be persisted into the refreshed note instead of discarded — or enrichment could be deferred until the summary phase settles for summary captures.

docs/plans/11-link-capture.md
```diff
@@ -107,7 +114,9 @@ Every capture lands in two phases so saving never waits on the network or AI:
 
 1. **Chrome extension** (`apps/extension`, MV3): action button + `⌘⇧P` (or `⌘⇧K` if reserved)
    to capture the active tab's URL, title, user-selected text/highlights, and a
-   screenshot (`captureVisibleTab`). Minimal UI: confirm + optional note. No keys, no AI.
+   screenshot (`captureVisibleTab`). Minimal UI: confirm + optional note and
+   independently selectable page text and on-device page summary (including both
+   together). No extension-held keys or provider-backed AI in this browser flow.
```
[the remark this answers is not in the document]
maccman: Fixed in 79ec07c4: Plan 11 now lists the complete version-1 envelope, including optional contentText, summary, and metaDescription fields.

packages/core/src/actions/capture-note.ts
```diff
@@ -102,6 +108,21 @@ export function capturePageTextFromBody(body: string): string | undefined {
...
+    (heading) => heading.level === 2 && heading.text.trim() === 'Summary',
+  )
+  if (!summaryHeading) {
+    return undefined
+  }
+  const summary = body
+    .slice(summaryHeading.to, sectionEnd(headings, summaryHeading, body.length))
+    .trim()
+  return summary === '' ? undefined : summary
+}
```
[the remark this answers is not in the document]
maccman: Fixed in 79ec07c4: generated summaries now have stable hidden start/end markers, extraction only reads the marked section, and a user-authored ## Summary collision is covered by a dedup regression test.

apps/extension/entrypoints/popup/capture-flow.ts
maccman: No code change here: the approved two-phase contract explicitly queues the raw link before extraction or generation and says closing during generation leaves that original capture intact. Page text and summary intentionally travel together in the second full update when both are selected; changing phase 1 to await extraction would give up the immediate durable link required by the plan.

apps/extension/entrypoints/popup/capture-flow.ts
maccman: Fixed in 84c1e03e: a same-day refresh now re-reads the existing capture, requires the note to remain non-private with a body hash matching captureHash, preserves unrelated frontmatter on a safe refresh, and consumes a stale update without touching user-owned content. Regression tests cover both edits and private: true.

apps/extension/lib/commands.ts
maccman: Fixed in 84c1e03e: instant capture is now used only when action.openPopup is absent. A present API that rejects propagates to the logged failure path and never silently captures; the rejected-openPopup branch is unit-tested.

apps/extension/entrypoints/popup/capture-flow.ts
maccman: Fixed in 84c1e03e: a rejected page-text fallback now returns update-rejected, the popup describes the generic capture update rejection, and the rejected fallback branch has a regression test.

apps/native-host/src/envelope.rs
maccman: No code change here: the approved delivery contract explicitly keeps this additive in version 1, co-releases extension and desktop host, and accepts that an older host may discard the optional field while the phase-1 link remains safe. A handshake or version bump would change that agreed compatibility contract.

apps/extension/entrypoints/popup/capture-flow.ts
maccman: Fixed by 22614ebf and strengthened in e8c2a9f5: dedup carries structured captureSummary, page text, and description forward even when the richer envelope sorts first. The same-millisecond summary-first regression test pins this ordering.

apps/extension/lib/page-summary.ts
maccman: Fixed in 84c1e03e: summary output uses safeParse and converts validation failure to a plain Error carrying the first issue message. The blank-output test now asserts the exact user-facing Error instead of substring-matching a ZodError blob.

apps/extension/entrypoints/background.ts
maccman: Fixed in 84c1e03e: every shortcut invocation now starts flushQueue before opening the popup or taking the legacy fallback, restoring the held-capture retry trigger even if the popup is dismissed.

packages/core/src/actions/capture-enrichment.ts
maccman: Fixed in 84c1e03e: after the provider returns, enrichment re-reads the latest still-managed snapshot and persists against its current captureHash, so a safe summary refresh retains the finished AI result. The regression now completes in one pass and asserts describePage was called exactly once.

apps/extension/lib/page-summary.ts
maccman: Fixed in 84c1e03e: both paragraph and word boundaries must now land in the latter half of the candidate; otherwise trimming uses a hard cut. A blob-heavy early-space regression test verifies the input no longer collapses to one word.

packages/core/src/actions/capture-drain.ts
```diff
@@ -232,16 +285,40 @@ export async function drainCaptureInbox(
...
+          continue
+        }
+      }
+      const mergedEnvelope = existing
+        ? {
+            ...envelope,
+            contentText: envelope.contentText ?? capturePageTextFromBody(existing.body),
+            summary: envelope.summary ?? existing.summary,
+            metaDescription: envelope.metaDescription ?? captureDescriptionFromBody(existing.body),
+          }
```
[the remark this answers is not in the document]
maccman: Fixed in 886bb330: capture notes now store a structured capturePageTextHash, and both dedup refreshes and BYOK enrichment recover only a marked Page Text block whose content matches that hash. User-authored marker text is never promoted or sent as page content; the collision regression covers both outcomes, while the reverse-order test continues to prove genuine page text survives.

maccman:
I think, in retrospect, might be better to do the summaries in the app with app linked llm.

PR #1165, opened 2026-08-22
feat: preview and edit web links in notes

maccman:
## Problem

Links in notes currently expose only their destination and Copy. Reflect cannot provide the title, description, or favicon needed by the richer Meowdown popup, and fetching that metadata from note-driven URLs creates privacy and local-network probing risks unless the native boundary enforces both note privacy and destination safety.

## Before -> After

| Surface | Before | After |
| --- | --- | --- |
| Hovered web link | URL and Copy | Favicon or globe, title, hostname, description, Copy/Edit/Remove, and title replacement |
| URL-labelled link | Raw URL remains the label | Replace URL with its title when the label still represents the destination |
| Edit form | Destination and hidden tooltip | Required Text and Link fields; authored CommonMark tooltip remains preserved |
| Private note | No metadata integration | Fails closed before every network phase |
| Local or intranet URL | Could be passed to a generic transport | Preview transport rejects loopback, private, link-local, and unsafe redirect targets |

## Changes

1. Note-scoped preview resolver
   - pass resolveLinkPreview from NotePane through NoteEditor into Meowdown
   - cache successful and failed results per editor session and deduplicate concurrent lookups
   - read the live full source through the open-document registry before every cache miss
   - parse frontmatter and mint the href through cloudSafeLinkHref; missing, unreadable, and private notes fail closed
   - re-check privacy and graph/session identity after page metadata, before favicon retrieval, and after favicon retrieval so stale work is discarded

2. Local metadata extraction
   - reuse the existing OpenGraph and HTML scraper with og:title then title, and og:description then the standard description
   - collapse whitespace, cap display values, resolve declared raster icons relative to the final page URL, and fall back to favicon.ico
   - keep failures silent and avoid persistent metadata storage, AI rewriting, and thumbnail fetching

3. Bounded native transport
   - extract the capture HTML request into a shared helper while retaining capture behavior
   - add validated Tauri commands for page HTML/final URL and normalized PNG favicon data
   - enforce HTTP(S), one total timeout across DNS and redirects, manual redirect validation, byte and MIME caps, decoder dimensions/allocation limits, and a no-proxy public-address pin
   - reject loopback, private-network, link-local, documentation, benchmark, and mixed public/private DNS answers before every request
   - normalize supported raster favicon formats to a small PNG data URL and reject SVG or malformed images

4. Meowdown feature dependency and styling
   - pin Core and React to immutable snapshot bdd909c from prosekit/meowdown#506
   - provide Reflect hover-card border, radius, and shadow tokens so link cards match wiki/backlink preview styling and motion

## Tests

- metadata precedence, whitespace normalization, display caps, relative icons, SVG fallback, and missing titles
- Zod contracts for bounded HTML/final URLs and PNG data URLs
- public-cloud href gating and private-note refusal
- resolver caching, concurrent deduplication, non-web refusal, missing/unreadable/private notes, privacy changes during each request phase, and stale graph/session work
- browser integration from NoteEditor into the Meowdown resolver in Chromium and WebKit
- favicon MIME/decoder normalization and public-network address enforcement in Rust

## Verification

- pnpm install --frozen-lockfile
- pnpm check
- targeted Vitest suite: 72 tests passed in Chromium/node
- targeted WebKit suite: 45 tests passed
- pnpm --filter @reflect/desktop sidecar
- cargo test -p reflect-open web_fetch: 5 passed
- cargo test -p reflect-open link_preview: 3 passed
- no additional visual or simulator pass was run, per request

## Risk / rollout

The dependency is intentionally pinned to the immutable Meowdown feature snapshot so this PR can build before a release. After Meowdown revert PR #505 and combined PR #506 are reviewed and merged by a maintainer, replace the snapshot with the published version in a follow-up before merging Reflect if that release is available.

Metadata requests remain local to the app. No note content is sent, private notes fail closed, no persistent preview cache is introduced, and no application version or release manifest is changed.

## Summary by CodeRabbit

- **New Features**
  - Added link previews for HTTP(S) links, including page titles, descriptions, site names, and favicon icons.
  - Added support for GIF, ICO, JPEG, PNG, and WebP favicon formats, normalized for display.
  - Added privacy-aware link handling for private notes and cloud-safe destinations.

- **Bug Fixes**
  - Improved retry behavior for temporary page and favicon fetch failures.
  - Rejects unsafe redirects, invalid content, oversized responses, and unsupported icons.
  - Improved metadata handling for incomplete or malformed pages.

image


apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+    let segments = address.segments();
+    !(address.is_unspecified()
+        || address.is_loopback()
+        || address.is_multicast()
+        || (segments[0] & 0xfe00) == 0xfc00
+        || (segments[0] & 0xffc0) == 0xfe80
+        || (segments[0] & 0xffc0) == 0xfec0
+        || (segments[0] == 0x2001 && segments[1] == 0x0db8))
+}
```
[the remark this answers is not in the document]
maccman: Fixed in f1dfd1f8. Public-network validation now rejects private IPv4 payloads in compatible, 6to4, and well-known NAT64 IPv6 forms, with public/private regression cases.

apps/desktop/src/editor/use-link-preview.ts
```diff
@@ -0,0 +1,132 @@
...
+      ...(iconSrc === null ? {} : { iconSrc }),
+    }
+  }
+
+  return (href) => {
+    const cached = cache.get(href)
+    if (cached !== undefined) return cached
+    const pending = resolve(href)
+    cache.set(href, pending)
+    return pending
```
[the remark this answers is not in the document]
maccman: Fixed in f1dfd1f8. Every resolver call re-reads live privacy before consulting the cache, and authorization denials are removed rather than cached. Tests cover public→private, private→public, and mid-resolution denial recovery.

packages/core/src/actions/meta-scrape.ts
```diff
@@ -48,6 +58,51 @@ export function parsePageMeta(html: string): PageMeta {
...
+export function parseLinkPreviewMeta(html: string, finalUrl: string): LinkPreviewMeta | null {
+  const document = new DOMParser().parseFromString(html, 'text/html')
+  const page = pageMetaFromDocument(document)
+  const title = clean(page.title)?.slice(0, MAX_LINK_TITLE_CHARS) ?? null
+  if (title === null) return null
+  return {
+    title,
+    description: clean(page.description)?.slice(0, MAX_LINK_DESCRIPTION_CHARS) ?? null,
+    iconUrl: rasterIconUrl(document, finalUrl),
+  }
```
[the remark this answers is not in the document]
maccman: Fixed in f1dfd1f8. LinkPreviewMeta now has a Zod-validated HTTP(S) boundary with bounded readonly fields; malformed and non-web final URLs return no metadata.

packages/core/src/exports/ai-actions.ts
```diff
@@ -58,7 +58,7 @@ export {
   type CloudSafe,
   type CloudSearchHit,
   type CloudSendable,
-} from '../ai/checkers'
+} from '../privacy/checkers'
```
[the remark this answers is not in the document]
maccman: Verified as not applicable. @reflect/core publishes only the package root and ./capture-envelope; there is no @reflect/core/ai-actions subpath, and repository search found no deep imports. The root @reflect/core surface still exports all three symbols through exports/link-preview.ts, so no supported consumer API changed.

packages/core/src/link-preview/metadata.ts
```diff
@@ -0,0 +1,93 @@
...
+    description:
+      metaContent(document, 'meta[property="og:description"]') ??
+      metaContent(document, 'meta[name="description"]'),
+    siteName: metaContent(document, 'meta[property="og:site_name"]'),
+  }
+}
+
+/** Extract {@link PageMeta} from an HTML document's text. Never throws. */
+export function parsePageMeta(html: string): PageMeta {
+  return pageMetaFromDocument(new DOMParser().parseFromString(html, 'text/html'))
```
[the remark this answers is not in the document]
maccman: Fixed in a74cd1c7. PageMeta now has the same bounded Zod validation as LinkPreviewMeta while preserving the never-throw parser contract.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,398 @@
...
+        ));
+    }
+    if segments[..6] == [0x0064, 0xff9b, 0, 0, 0, 0] {
+        return is_public_ipv4(Ipv4Addr::new(
+            (segments[6] >> 8) as u8,
+            segments[6] as u8,
+            (segments[7] >> 8) as u8,
+            segments[7] as u8,
+        ));
+    }
```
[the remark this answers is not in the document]
maccman: Fixed in 610a944a. The full 64:ff9b:1::/48 local-use translation prefix is rejected, with a regression case in the non-public range table.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+async fn fetch_html(value: &str, scope: NetworkScope) -> AppResult<FetchResponse> {
+    let response = fetch(
+        value,
+        scope,
+        "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+        HTML_MAX_BYTES,
+        LimitBehavior::Truncate,
+    )
+    .await?;
+    if !response.content_type.contains("html") {
```
maccman: **Content-Type is checked only after the whole body is streamed.** `fetch()` drains up to `max_bytes` (2 MiB HTML / 512 KiB icon) before `fetch_html` rejects non-HTML and before `link_preview_fetch_icon` rejects non-raster types. The old `capture_meta_fetch` checked headers before reading any body, so capture inherits the regression too.

Hovering a link to a PDF/MP4/zip now downloads 2 MiB just to throw it away; an `image/svg+xml` favicon pulls 512 KiB before `is_raster_content_type` refuses it. Suggest passing a `content_type_ok: fn(&str) -> bool` (or the expected MIME family) into `fetch()` and bailing right after `send()` on the headers, before the chunk loop.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+            let location = response
+                .headers()
+                .get(reqwest::header::LOCATION)
+                .and_then(|value| value.to_str().ok())
+                .ok_or_else(|| AppError::parse(format!("redirect from {url} has no Location")))?;
+            url = url.join(location).map_err(|error| {
+                AppError::parse(format!("invalid redirect from {url}: {error}"))
+            })?;
+            url = parse_http_url(url.as_str())?;
```
maccman: **Redirects may downgrade https → http.** Each hop is re-validated only via `parse_http_url`, which accepts both schemes, so an `https://` link that 302s to `http://…` is followed in cleartext with the browser-mimicking header set. reqwest's old `Policy::limited(5)` allowed this for capture too, so it isn't a capture regression — but the `PublicHttp` editor path is new surface and is specifically about not letting note-driven URLs reach unexpected places. Consider rejecting a hop whose scheme goes from `https` to `http` (at least in `PublicHttp` scope).

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+async fn client_for(
+    url: &Url,
+    scope: NetworkScope,
+    deadline: Instant,
+) -> AppResult<reqwest::Client> {
+    let mut builder = reqwest::Client::builder()
+        .redirect(redirect::Policy::none())
+        .timeout(remaining_timeout(deadline)?)
```
maccman: **Request timeout is computed before DNS runs.** `.timeout(remaining_timeout(deadline)?)` is evaluated here, then `resolve_public` below spends part of the budget on DNS, so each hop's HTTP phase gets the pre-DNS remainder and can overrun the 15 s deadline by the DNS duration on every hop. Move the `.timeout(...)` after `resolve_public` (or apply the per-request deadline via `RequestBuilder::timeout` at send time).

Also note the `spawn_blocking` resolve task is not cancelled when `tokio::time::timeout` fires, so a stalled resolver leaves an orphaned blocking-pool thread per timed-out hover until getaddrinfo gives up on its own.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+        .host_str()
+        .ok_or_else(|| AppError::parse(format!("URL has no host: {url}")))?
+        .to_owned();
+    let port = url
+        .port_or_known_default()
+        .ok_or_else(|| AppError::parse(format!("URL has no port: {url}")))?;
+    let resolve_host = host.clone();
+    let resolve_task = tauri::async_runtime::spawn_blocking(move || {
+        (resolve_host.as_str(), port)
+            .to_socket_addrs()
```
maccman: **IPv6-literal hosts can never resolve in `PublicHttp` scope.** `Url::host_str()` returns the bracketed form (`[2606:4700:4700::1111]`) and `(&str, u16).to_socket_addrs()` treats that as a hostname (std only parses bare IPv6 text), so it fails with "nodename nor servname provided". `resolve_to_addrs` is likewise keyed on the bracketed string. The unit tests only pass because they `trim_matches(['[', ']'])` before building the `SocketAddr`, so the production path is untested. Fail-closed, but a public IPv6-literal link silently never gets a preview. Use `url.host()` and short-circuit `Host::Ipv4`/`Host::Ipv6` literals instead of going through `to_socket_addrs`.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+) -> AppResult<FetchResponse> {
+    let deadline = Instant::now() + FETCH_TIMEOUT;
+    let mut url = parse_http_url(value)?;
+    for redirect_count in 0..=MAX_REDIRECTS {
+        let client = client_for(&url, scope, deadline).await?;
+        let response = client
+            .get(url.clone())
+            .header("Accept", accept)
+            .header("Accept-Language", "en-US,en;q=0.9")
+            .header("Sec-Fetch-Dest", "document")
```
maccman: **Favicon fetches present as top-level navigations.** These `Sec-Fetch-Dest: document` / `Sec-Fetch-Mode: navigate` / `Upgrade-Insecure-Requests` headers are hardcoded for every caller, so `link_preview_fetch_icon` sends a navigation profile with an image-only `Accept` — a signature no real browser emits (an icon request would be `Sec-Fetch-Dest: image`, `Sec-Fetch-Mode: no-cors`). Fingerprinting CDNs may 403/challenge it, and the resulting globe fallback is then cached for the session. Let the caller supply its request profile alongside `accept`.

apps/desktop/src-tauri/src/web_fetch.rs
```diff
@@ -0,0 +1,402 @@
...
+        })?
+        .map_err(|error| AppError::io(format!("DNS task failed: {error}")))?
+        .map_err(|error| AppError::Network {
+            message: format!("could not resolve {host}: {error}"),
+        })?;
+    validate_public_addresses(url, &addresses)?;
+    Ok((host, addresses))
+}
+
+async fn client_for(
```
maccman: **A new `reqwest::Client` and a fresh DNS lookup per redirect hop.** `Client::build` loads the TLS config/root store each time and discards the connection pool, and `resolve_public` does a blocking `getaddrinfo` per hop; the page fetch and the favicon fetch to the same host then resolve independently again. A link that 301s twice plus a favicon with one redirect = 5 client builds and 5 lookups for one host. A process-wide client (`OnceLock`) with a custom `dns_resolver` that runs `validate_public_addresses` would keep the SSRF pin, reuse the pool, and let you cache validated `(host, port)` for the duration of one resolve.

apps/desktop/src/editor/link-preview-resolver.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    // Authorization is deliberately outside the cache: a completed preview
+    // must disappear as soon as its note becomes private, while a denied
+    // lookup must become eligible when the note becomes public again.
+    const pageUrl = await authorizeOutboundUrl(href)
+    if (pageUrl === null) return
+
+    let pending = cache.get(href)
+    if (pending === undefined) {
+      pending = resolve(href, pageUrl)
```
maccman, line 164: **Transient failures are cached for the whole editor session.** Every `fetchHtml` rejection becomes `FAILED_RESOLUTION` and is stored here with no TTL; only `authorization-denied` is evicted (below). The Rust side deliberately distinguishes retryable `Network` (429/5xx/timeout) from permanent `io` via `classify_fetch_status`, and that kind is serialized (`isAppError(e).kind === 'network'`), but the bare `catch {}` in `resolve()` discards it.

Concretely: hover a link while briefly offline, or get one 429, and the daily stream keeps that pane mounted (`keepMounted`), so the link never shows a preview again until the pane unmounts or the graph generation changes. Suggest not caching `network`-kind failures (or giving failures a short TTL).

apps/desktop/src/editor/link-preview-resolver.ts
```diff
@@ -0,0 +1,151 @@
...
+ * successful and failed lookups and naturally deduplicates concurrent calls.
+ */
+export function createNoteLinkPreviewResolver(
+  session: LinkPreviewSession,
+  currentSession: () => LinkPreviewSession | null,
+  dependencies: LinkPreviewResolverDependencies = defaultDependencies,
+): LinkPreviewResolver {
+  const cache = new Map<string, Promise<LinkPreviewResolution>>()
+
+  async function authorizeOutboundUrl(url: string): Promise<CloudSafe<string> | null> {
```
maccman: **Four full-note disk reads + frontmatter parses per first preview, one per cached hover.** `authorizeOutboundUrl` runs at entry, after metadata, before the icon fetch, and after it; each call goes through `readExistingNoteSource`, which always does a `readNote` IPC (even when the live buffer is used — the read is a liveness probe) and then `splitFrontmatter`/`parseFrontmatter` over the entire note text to read one boolean. The two pre-icon calls (`authorizeOutboundUrl(href)` then `authorizeOutboundUrl(metadata.iconUrl)`) are back-to-back with no network work between them, so at minimum those can share one `readSource` snapshot. Better: derive `isPrivate` once per session/buffer revision and only re-read when the buffer changes.

apps/desktop/src/editor/link-preview-resolver.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (!sameSession(currentSession(), session)) return null
+    try {
+      const source = await dependencies.readSource(session.path, session.generation)
+      if (!sameSession(currentSession(), session)) return null
+      const frontmatter = parseFrontmatter(splitFrontmatter(source).raw).data
```
maccman: nit (reuse): this is byte-for-byte `notePrivate()` from `packages/core/src/actions/capture-note.ts:146` (and the same expression again in `ai/chat/tools.ts:161`). It just isn't re-exported from `@reflect/core` yet — a one-liner in `exports/link-preview.ts` would let the three privacy gates share one definition so they can't drift.

apps/desktop/src-tauri/src/capture.rs
```diff
@@ -645,8 +554,8 @@ pub async fn capture_oembed_fetch(url: String) -> AppResult<String> {
     }
     let client = reqwest::Client::builder()
         .redirect(reqwest::redirect::Policy::none())
-        .timeout(META_FETCH_TIMEOUT)
-        .user_agent(META_FETCH_USER_AGENT)
+        .timeout(FETCH_TIMEOUT)
```
maccman: nit (reuse): `capture_oembed_fetch` still hand-rolls the client build, status classification, and reject-on-overflow chunk loop that `web_fetch::fetch(..., LimitBehavior::Reject)` now provides — it even imports `classify_fetch_error`/`classify_fetch_status`/`FETCH_TIMEOUT`/`USER_AGENT` to do so. The only missing knob is "no redirects" (and the https-only pre-check can stay here). Worth folding in so the module doc's "shared by capture" claim holds and the two transports don't drift.

maccman:
Review summary (10 inline comments posted, on head 610a944).

**Verified locally:** `pnpm install --frozen-lockfile` pulls the bdd909c snapshot; `tsc -b` passes, `cargo check --tests` compiles, and the new vitest/Rust suites pass. The `--meowdown-hover-card-*` tokens and `resolveLinkPreview` both exist in the pinned build.

**Most important inline findings** (all in `web_fetch.rs` / `link-preview-resolver.ts`):
1. Content-Type is gated only after the full 2 MiB / 512 KiB body is streamed — capture inherits this regression.
2. Redirect loop allows https→http downgrade on the new public-only path.
3. Transient failures (429/5xx/offline) are cached for the whole editor session; the Rust retryable/permanent distinction is discarded.
4. Per-hop request timeout is computed before DNS, so hops can overrun the 15 s deadline.
5. Bracketed IPv6-literal hosts never resolve in `PublicHttp` scope (tests strip the brackets).

**Not inline, worth noting:**
- The pkg.pr.new pin is acknowledged in the description as a merge gate; just flagging that `packages/core` still resolves `@meowdown/markdown@0.66.0` while the editor gets the bdd909c copy, so the lockfile carries two markdown instances until the published release replaces the snapshot.
- `ai/checkers.ts` → `privacy/checkers.ts` left stale paths in `docs/porting/README.md`, `docs/porting/ai-menu-and-prompts.md`, `docs/plans/10-ai-copilot-sidebar.md`, `docs/plans/architecture-conventions.md`; `docs/security-pass/findings.md` still says the capture buffer is 512 KB.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in 0ce5c0f2 and retained in e5eda38b. Public IPv4 and IPv6 literals are parsed and validated directly before the request and never sent through DNS. The production path is covered by accepts_public_ipv6_literals_without_dns.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in 0ce5c0f2 and e5eda38b. Each transport now supplies a content-type validator, and fetch runs it immediately after status/header parsing, before entering the body chunk loop. HTML, JSON, and raster icon responses therefore reject without draining an unsuitable body.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in e5eda38b. The remaining overall deadline is applied on RequestBuilder immediately before send, so it covers DNS, connect, response headers, and body streaming. Public DNS now runs behind the shared client resolver; because OS getaddrinfo cannot be forcibly cancelled once running, concurrent blocking lookups are capped at eight so timed-out hovers cannot grow the blocking pool without bound.

apps/desktop/src/editor/link-preview-resolver.ts
maccman: Fixed in e5eda38b. Network-kind page failures evict the pending cache entry and retry on the next hover. A network-kind favicon failure still returns the metadata/globe preview but evicts it so the icon can retry. Permanent failures remain session-cached. Both paths have tests.

apps/desktop/src-tauri/src/capture.rs
maccman: Fixed in e5eda38b. capture_oembed_fetch now delegates to fetch_capture_json, which preserves HTTPS-only, JSON MIME validation before body streaming, the 64 KiB reject-on-overflow cap, and zero redirects.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in 0ce5c0f2. PublicHttp redirects now reject an HTTPS to HTTP downgrade at every hop. The capture scope preserves its existing HTTP redirect behavior.

apps/desktop/src/editor/link-preview-resolver.ts
maccman: Fixed in e5eda38b. The post-metadata page recheck and pre-favicon authorization now share one live source read and mint both URLs from that snapshot. The entry and post-favicon reads remain deliberate fail-closed checkpoints, and cached previews still perform one live read because the open-document API currently exposes no buffer revision/liveness token that could safely replace it.

apps/desktop/src/editor/link-preview-resolver.ts
maccman: Fixed in e5eda38b. notePrivate now lives in privacy/checkers, is re-exported by the link-preview barrel, and is shared by the resolver and chat privacy probe. capture-note re-exports it to preserve its existing internal consumers.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in e5eda38b. Fetch callers now select a request profile. Favicon requests send Sec-Fetch-Dest: image and Sec-Fetch-Mode: no-cors, and omit the navigation-only site and upgrade headers.

apps/desktop/src-tauri/src/web_fetch.rs
maccman: Fixed in e5eda38b. Each network scope now owns one process-wide reqwest client. Public requests use a no-proxy custom resolver that validates every DNS answer before returning pinned addresses to the connector, while the shared connection pool is reused across redirects, page metadata, and favicon requests.

ocavue (approved):

PR #1192, opened 2026-09-01
fix: improve iOS note deletion

maccman:
## Problem

iOS note deletion had two rough edges:

1. The confirmation appeared as a centered, desktop-style dialog after both the note actions menu and swipe-to-delete flow, breaking the mobile bottom-sheet interaction pattern.
2. A newly created note is intentionally buffer-only until its first meaningful edit is saved. Deleting it while the initial read or first write was still pending could call the trash command before the file existed, surfacing “no such file or directory” or allowing a late write to recreate the note.

## Before -> After

| Before | After |
| --- | --- |
| Centered delete dialog | Bottom-anchored drawer that slides up from the screen edge |
| Dialog footer actions | Full-width destructive and cancel actions sized for touch |
| Deleting during initial load or the first save could race the filesystem | Deletion pauses new saves and waits for loading/started writes before choosing local discard or trash |
| Failed filesystem deletion could leave persistence paused | The live session resumes saving after a failed trash operation |

## Changes

1. Replace the shared `NoteDeleteDialog` with `NoteDeleteDrawer`, built on the existing mobile `Drawer` primitive.
2. Keep the recovery explanation, error state, no-graph guard, in-flight dismissal lock, and duplicate-submit protection unchanged.
3. Route both note-screen deletion and list swipe deletion through the new drawer.
4. Add the session-owned `prepareDelete()` / `cancelDelete()` lifecycle. Preparation pauses new saves, waits for the initial load and any write already in flight, then reports whether a backing file exists.
5. Update `deleteOpenNote()` to discard a still-unpersisted session locally or trash its settled file. A failed trash operation resumes normal persistence; a successful deletion discards the prepared session so it cannot recreate the note.
6. Exercise the real lazy `NoteSession` in regression tests rather than mocking its persistence classification.

## Tests

- Note deletion: deleting immediately while the lazy note is still loading never calls the filesystem.
- Initial-write race: the filesystem delete is held until a deliberately blocked first write lands, then removes the created file without an error or resurrection.
- Failure recovery: a prepared session resumes its pending save after trashing fails.
- Persisted-note and daily-note behavior remains covered, along with drawer presentation and mobile entry points.
- Manual iPhone-sized browser check: drawer opening, visual presentation, and cancellation without deletion.

## Verification

- `pnpm test --run apps/desktop/src/mobile/note-delete.test.ts apps/desktop/src/lib/note-delete.test.ts apps/desktop/src/editor/note-session.test.ts apps/desktop/src/editor/document-binding.test.ts apps/desktop/src/editor/open-documents.test.ts apps/desktop/src/editor/move-note.test.ts apps/desktop/src/editor/rename-coordinator.test.ts` — 107 tests passed
- `pnpm test --run apps/desktop/src/mobile/note-actions-menu.test.tsx apps/desktop/src/mobile/note-row-list.test.tsx` — 17 tests passed
- `pnpm test --run apps/desktop/src/mobile/mobile-screen.test.tsx` — 29 tests passed
- `pnpm check`
- `pnpm build`

## Risk / Rollout

No schema, migration, or backend delete-command changes. Coordination is confined to an open note’s existing persistence state machine: scheduled saves pause during deletion, already-started work settles, and failed filesystem deletion explicitly resumes persistence. Notes without an open session retain the existing trash path.

## Summary by CodeRabbit

- **New Features**
  - Added a shared mobile confirmation drawer for deleting notes.
  - Deletion now provides progress handling and error feedback.
  - Cancel and delete actions are disabled while deletion is in progress.

- **Bug Fixes**
  - Updated mobile note actions and note lists to use the consistent drawer experience.
  - Improved handling of failed deletions while keeping confirmation available.
  - Unpersisted notes are discarded locally without attempting file deletion.
  - Deletion now waits for pending saves to complete before removing persisted notes.


ocavue:
I've tested this PR.  I can still see the “no such file or directory” error.

Before:

image image


After:

image image

apps/desktop/src/mobile/note-delete.test.ts
```diff
@@ -28,9 +30,18 @@ afterEach(() => {
 })
 
 describe('deleteOpenNote', () => {
+  it('discards an unpersisted note without deleting a missing file', async () => {
+    openSessionMock.mockReturnValue({ isUnpersisted: () => true, discard })
```
ocavue: Try not to mock `isUnpersisted`. The logic inside `isUnpersisted` is what we want to test.

maccman: Fixed in a02a8872. The mobile regression now uses a real NoteSession and covers deletion during both initial loading and an in-flight first write.

maccman:
Ocavue’s reproduced “no such file or directory” race is fixed in a02a8872. Deletion now pauses new saves, waits for initial loading and any started first write, then re-evaluates local discard versus filesystem trash. The regression test uses a real NoteSession for both timing windows.

PR #1205, opened 2026-09-02
fix: recover interrupted app-owned merges

LemieuxStrategicSolutions:
## Problem
If the process is terminated after libgit2 enters merge state but before Reflect commits and cleans up, the repository remains in `Merge` state. Every later sync fails the clean-state guard, so the device stays at Needs attention until manual Git repair.

This was reproduced on a physical iPhone, including a stale zero-byte `.git/index.lock`.

## Solution
- write a versioned, app-owned recovery marker before beginning a non-fast-forward merge
- store the protected local and fetched remote OIDs
- on the next sync, recover only when that exact marker proves the interrupted merge belongs to Reflect
- remove a stale index lock, restore the committed local parent, and deterministically replay the fetched merge
- leave foreign CLI merges untouched
- recognize the case where the merge commit reached HEAD but cleanup was interrupted

The marker lives under `.git`, is per-device, and never enters the graph.

## Validation
- `cargo fmt --check`
- `cargo test git::tests::` — 34 passed
- tests cover interrupted app merge with stale lock and foreign merge preservation

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved merge recovery after interruptions, automatically rolling back and replaying incomplete app-managed merges during the next sync.
  - Added safeguards to prevent interrupted or failed merges from leaving the repository in an unusable state.
  - External merge operations are now preserved and clearly reported instead of being modified automatically.
  - Merge state is cleaned up after successful completion, reducing stale merge conditions.

### DIFF
--- a/apps/desktop/src-tauri/src/git/merge.rs
+++ b/apps/desktop/src-tauri/src/git/merge.rs
@@ -23,8 +23,8 @@ use std::fs;
 use std::path::Path;
 
 use git2::build::CheckoutBuilder;
-use git2::{Index, IndexEntry, MergeOptions, Repository};
-use serde::Serialize;
+use git2::{Index, IndexEntry, MergeOptions, Oid, Repository, ResetType};
+use serde::{Deserialize, Serialize};
 
 use crate::error::AppResult;
 
@@ -34,6 +34,21 @@ use super::repo::{current_branch, ensure_clean_state, open_existing, signature};
 /// the remote one — product language, not branch names.
 const OUR_LABEL: &str = "this device";
 const THEIR_LABEL: &str = "other device";
+
+/// Written before libgit2 enters merge state and removed only after the merge
+/// commit (or a verified rollback) completes. If iOS terminates the process in
+/// that window, the next sync can prove the operation belongs to Reflect,
+/// restore local HEAD, and replay the fetched merge. Foreign CLI merges remain
+/// hands-off.
+const MERGE_RECOVERY_FILE: &str = "REFLECT_MERGE_STATE";
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct MergeRecovery {
+    version: u32,
+    local_oid: String,
+    remote_oid: String,
+}
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
@@ -98,6 +113,9 @@ fn side_of(entry: Option<IndexEntry>) -> Option<ConflictSide> {
 /// (the sync engine guarantees it): local changes are already committed.
 pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
     let repo = open_existing(root)?;
+    if let Some(outcome) = recover_interrupted_merge(&repo, root)? {
+        return Ok(outcome);
+    }
     ensure_clean_state(&repo)?;
     let branch = current_branch(&repo)?;
     let Ok(remote_oid) = repo.refname_to_id(&format!("refs/remotes/origin/{branch}")) else {
@@ -142,14 +160,12 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
         });
     }
 
-    let mut merge_opts = MergeOptions::new();
-    let mut checkout = CheckoutBuilder::new();
-    checkout
-        .allow_conflicts(true)
-        .conflict_style_merge(true)
-        .our_label(OUR_LABEL)
-        .their_label(THEIR_LABEL);
-    repo.merge(&[&annotated], Some(&mut merge_opts), Some(&mut checkout))?;
+    let local_oid = repo.head()?.peel_to_commit()?.id();
+    write_recovery_marker(&repo, local_oid, remote_oid)?;
+    if let Err(error) = begin_merge(&repo, &annotated) {
+        let _ = rollback_owned_merge(&repo, local_oid);
+        return Err(error);
+    }
 
     // From here the repo carries MERGE_* state; a failure that leaves it
     // behind would trip `ensure_clean_state` on every later cycle and wedge
@@ -157,9 +173,10 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
     // on every path; the next cycle re-derives anything a failed attempt lost.
     let result = complete_merge(&repo, root, remote_oid);
     if result.is_err() {
-        let _ = repo.cleanup_state();
+        let _ = rollback_owned_merge(&repo, local_oid);
     }
     let (conflicted_paths, changed_files) = result?;
+    remove_recovery_marker(&repo)?;
 
     let kind = if conflicted_paths.is_empty() {
         MergeKind::Merged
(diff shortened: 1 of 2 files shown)


apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -98,6 +113,9 @@ fn side_of(entry: Option<IndexEntry>) -> Option<ConflictSide> {
 /// (the sync engine guarantees it): local changes are already committed.
 pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
     let repo = open_existing(root)?;
+    if let Some(outcome) = recover_interrupted_merge(&repo, root)? {
```
maccman, line 116: **Recovery is unreachable from the real sync cycle.**

Every engine cycle (`packages/core/src/sync/engine.ts` `cycle()`) runs `gitCommitAll` *before* `gitFetch`/`gitMergeRemote`, and `commit_all` has its own `ensure_clean_state` (`commit.rs:49`) plus an unconditional `index.write()`. With `MERGE_HEAD` and the zero-byte `index.lock` this PR targets, `commit_all` fails with "the backup repository has a Merge in progress" (or "the index is locked"), the cycle throws, and `merge_remote` is never called. The marker is never read; the device stays at Needs attention exactly as before.

I probed this: after `interrupt_next_merge_for_test`, `commit_all(root_a, "Update notes", MAX_FILE_BYTES)` returns that `Io` error. The new test only passes because it calls `merge_remote` directly.

The recovery needs to live where the guard lives — e.g. inside `ensure_clean_state` or a shared "open graph repo for sync" step in `repo.rs` — so `commit_all`, `merge_remote`, and `status` all self-heal.

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(()),
+        Err(error) => Err(error.into()),
+    }
+}
+
+fn rollback_owned_merge(repo: &Repository, local_oid: Oid) -> AppResult<()> {
+    remove_stale_index_lock(repo)?;
+    repo.cleanup_state()?;
+    let local = repo.find_object(local_oid, None)?;
+    repo.reset(&local, ResetType::Hard, None)?;
```
maccman, line 323: **The hard reset discards uncommitted note edits.**

`rollback_owned_merge` runs on every failure and recovery path and rewrites every tracked file to `local_oid`. The "working tree is committed" precondition only holds at `commit_all` time; the editor keeps saving during `gitFetch` (seconds), and after a crash the user may type before the launch cycle fires. Pre-PR, the `cleanup_state`-only path left those files dirty and the next cycle committed them.

Probe: edit `notes/local.md` after `interrupt_next_merge_for_test`, call `merge_remote` → the file is back to `# This device\n`; the edit is gone and the watcher reloads the reverted content into the editor. The same happens when `begin_merge` fails with `GIT_ECONFLICT` ("uncommitted change would be overwritten"), which is reachable via an oversized tracked file `commit_all` deliberately leaves uncommitted.

Note libgit2's `git_merge` already runs `merge_state_cleanup` on its own errors, so the `begin_merge` failure path mostly needs only the marker removed. For the recovery path, either scope the checkout to paths the merge touched, or snapshot files that differ from both `local_oid` and the merge result before resetting.

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+        let mut changed_files = changed_between(repo, Some(&local_tree), &merged_tree)?;
+        stamp_modified_times(root, &mut changed_files);
+        return Ok(Some(MergeOutcome {
+            kind: MergeKind::Merged,
+            conflicted_paths: Vec::new(),
+            changed_files,
+        }));
+    }
+
+    if head_oid != local_oid {
```
maccman, line 233: **This is a permanent, silent wedge with no self-heal path.**

Nothing ever removes the marker on this branch, so once HEAD moves the error recurs on every full cycle. `commit_all` keeps succeeding (state is Clean, and no UI reads `inProgress`), so edits pile up unpushed with an error naming a file the user cannot know about.

Realistic path: crash mid-merge → user follows the existing "finish or abort it with git first" advice and runs `git merge --abort` → edits a note → `commit_all` advances HEAD → this line fires forever. Also reached by a kill between `cleanup_state` (line 373) and `remove_recovery_marker` (line 179) followed by any local commit. Probe: three consecutive `merge_remote` calls all return this error with the marker still on disk.

The same applies to an unparseable or unsupported-version marker at lines 273–281.

A Clean repo whose HEAD descends from `local_oid` (or from a `(local, remote)` merge commit) proves the marker is stale: delete it and continue rather than error.

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -142,24 +160,23 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
     // behind would trip `ensure_clean_state` on every later cycle and wedge
     // sync until a manual repair — exactly what this design forbids. Clear it
     // on every path; the next cycle re-derives anything a failed attempt lost.
     let result = complete_merge(&repo, root, remote_oid);
     if result.is_err() {
-        let _ = repo.cleanup_state();
+        let _ = rollback_owned_merge(&repo, local_oid);
     }
     let (conflicted_paths, changed_files) = result?;
+    remove_recovery_marker(&repo)?;
```
maccman, line 179: **A marker-unlink failure fails a merge that already committed.**

At this point the merge commit is on HEAD and `cleanup_state` has run. If the unlink fails (EACCES/EBUSY/transient IO), `merge_remote` returns `Err`: `engine.ts` `step()` never runs `onResult` on rejection, so `changedFiles` are not fanned out for reindex and the engine emits an error. Worse, the user keeps editing, a debounce `push` cycle commits on top of the merge commit, and the next full cycle hits the permanent error at line 233.

Pre-PR the only post-commit step was the idempotent `cleanup_state`. Since a 2-parent commit on HEAD already makes the marker redundant, this should be best-effort (`let _ = …` plus a `tracing::warn!`).

maccman (commented):
Reviewed at 3e1ea93. Ten inline comments; the ones that matter most:

1. **The recovery is not reachable from the sync engine.** `cycle()` runs `gitCommitAll` first, and `commit_all` has its own `ensure_clean_state` + `index.write()`, both of which fail on the Merge state / stale `index.lock` this PR is meant to survive. So the iPhone repro from the description still wedges, one step earlier than `merge_remote`. The fix belongs at the guard (`ensure_clean_state` / a shared repo-open step), not inside `merge_remote` alone, and the new test should drive `commit_all → fetch → merge_remote` in engine order.
2. **`rollback_owned_merge`'s hard reset loses uncommitted note edits** made during the fetch window or after a crash. Pre-PR those stayed dirty and got committed next cycle.
3. **Several marker states become permanent wedges** (HEAD moved past `local_oid`, corrupt marker, marker-unlink failure after a landed commit). None are removed automatically and the error names a file the user cannot know about. A Clean repo whose HEAD descends from the marker's commits should just drop the marker.
4. The Merge-state arm should require `MERGE_HEAD == remote_oid` before rolling back, or it hard-resets a foreign CLI merge.

Two minor notes with no inline comment: the "commit reached HEAD" recovery branch always reports `Merged` with empty `conflictedPaths` even when the landed commit was the "(conflicts to review)" variant — nothing consumes that today, but worth reconstructing from the commit message. And per CONTRIBUTING.md the `fix:` title becomes the changelog line, so something like `fix: sync no longer gets stuck after the app is killed mid-merge` reads better than "app-owned merges".

Verified with clippy (clean) and by adding temporary probe tests against this branch; all probes have been reverted.

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+    }
+
+    if head_oid != local_oid {
+        return Err(crate::error::AppError::io(
+            "Reflect found an interrupted merge marker, but HEAD no longer matches its protected local commit; no automatic recovery was attempted",
+        ));
+    }
+
+    match repo.state() {
+        git2::RepositoryState::Clean | git2::RepositoryState::Merge => {
```
maccman, line 240: **This arm trusts the marker alone and never checks `MERGE_HEAD`.**

A stale marker at the same HEAD plus a later *foreign* CLI merge gets `cleanup_state`'d, hard-reset, and replaced by Reflect's replayed merge — the opposite of the "foreign CLI merges remain hands-off" promise in the marker's doc comment.

Probe: crash → `git merge --abort` (marker survives) → `git merge topic` mid-resolution (state Merge, `MERGE_HEAD` = topic, HEAD == `local_oid`) → `merge_remote` returns `Merged`; afterwards state is Clean, `notes/topic.md` is gone, `MERGE_HEAD` is gone, and HEAD is a fresh `(local, origin/main)` merge.

Today `commit_all`'s guard shields this in the engine, but fixing the ordering issue at line 116 removes that shield. Require `.git/MERGE_HEAD == recovery.remote_oid` (and `ORIG_HEAD == local_oid`) before rolling back; otherwise fall through to `ensure_clean_state`. (Sibling of CodeRabbit's point on line 217.)

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+    local_oid: Oid,
+    remote_oid: Oid,
+) -> AppResult<bool> {
+    let commit = repo.find_commit(head_oid)?;
+    Ok(commit.parent_count() == 2
+        && commit.parent_ids().any(|oid| oid == local_oid)
+        && commit.parent_ids().any(|oid| oid == remote_oid))
+}
+
+fn remove_stale_index_lock(repo: &Repository) -> AppResult<()> {
```
maccman, line 301: **Stale-lock removal only runs when a marker exists, but the lock is left by interruptions that never write one.**

A zero-byte `index.lock` is the artifact of *any* killed libgit2 index write: `commit_all`'s `index.write()` and the fast-forward `checkout_head` both leave it, and neither writes a marker. Probe with a zero-byte `.git/index.lock` and no marker:

- `commit_all` fails with "the index is locked; this might be due to a concurrent or crashed process" even on a clean tree — and since it runs first in every cycle, that is a permanent wedge.
- Fast-forward `merge_remote` moves `refs/heads/main` and writes `notes/remote.md`, then fails on the index write; the next call reports `UpToDate`, so the stale index and the lock persist forever.

Conversely, the removal here is unconditional whenever a marker exists, so a live lock held by a concurrent `git` process on desktop would be deleted. Plan 12 already scopes "stale-`index.lock` recovery on startup"; a staleness check at repo open (zero-byte, or mtime older than this process start) would cover every path and stay independent of the marker.

apps/desktop/src-tauri/src/git/tests.rs
```diff
@@ -780,6 +780,78 @@ fn failed_merge_completion_still_clears_the_merge_state() {
...
+
+    interrupt_next_merge_for_test(root_a).unwrap();
+    let repo = Repository::open(root_a).unwrap();
+    assert_eq!(repo.state(), git2::RepositoryState::Merge);
+    assert!(repo.path().join("REFLECT_MERGE_STATE").is_file());
+    // Model the zero-byte lock left by the physical iPhone interruption.
+    fs::write(repo.path().join("index.lock"), []).unwrap();
+    drop(repo);
+
+    let merged = merge_remote(root_a).unwrap();
```
maccman, line 808: **The test bypasses the engine's real ordering, and two advertised branches have no test.**

Inserting the engine's actual first step here —

```rust
commit_all(root_a, "Update notes", MAX_FILE_BYTES).unwrap();
```

— makes this test fail with "Merge in progress" (and it would also trip on the `index.lock` planted at line 805). The "merge commit reached HEAD but cleanup was interrupted" branch (`merge.rs:217`) and the HEAD-mismatch refusal (`merge.rs:233`) from the PR description are not exercised at all.

Worth adding: (a) `commit_all` → `fetch` → `merge_remote` from the interrupted state, asserting the graph converges and pushes; (b) marker left behind after a landed merge commit; (c) marker present with HEAD moved, asserting the marker is dropped rather than erroring.

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+}
+
+fn write_recovery_marker(repo: &Repository, local_oid: Oid, remote_oid: Oid) -> AppResult<()> {
+    let marker = MergeRecovery {
+        version: 1,
+        local_oid: local_oid.to_string(),
+        remote_oid: remote_oid.to_string(),
+    };
+    let path = repo.path().join(MERGE_RECOVERY_FILE);
+    let temporary = repo.path().join(format!("{MERGE_RECOVERY_FILE}.tmp"));
```
maccman, line 255: **Reuse `crate::capture::atomic_write_to` instead of a hand-rolled tmp+rename.**

`capture.rs:63` already does `NamedTempFile::new_in(dir)` + `write_all` + `flush` + `persist(path)` and is consumed cross-module (`skill.rs:19`). This version uses a fixed sibling name with no flush, so a kill between `fs::write` and `fs::rename` — the exact window this PR is about — strands `REFLECT_MERGE_STATE.tmp` in `.git` forever.

```suggestion
    crate::capture::atomic_write_to(
        &path,
        &serde_json::to_string(&marker).map_err(|error| {
            crate::error::AppError::io(format!("failed to encode merge recovery marker: {error}"))
        })?,
    )?;
    Ok(())
}
```

(Adjust the surrounding lines: drop `temporary`, `fs::write`, and `fs::rename`.)

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -173,6 +190,152 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
+fn rollback_owned_merge(repo: &Repository, local_oid: Oid) -> AppResult<()> {
+    remove_stale_index_lock(repo)?;
+    repo.cleanup_state()?;
+    let local = repo.find_object(local_oid, None)?;
+    repo.reset(&local, ResetType::Hard, None)?;
+    remove_recovery_marker(repo)
+}
+
+#[cfg(test)]
+pub(super) fn interrupt_next_merge_for_test(root: &Path) -> AppResult<()> {
```
maccman, line 328: **Duplicates `merge_remote`'s preamble and is the only `pub(super) fn` in `git/` without a doc comment.**

The open / clean-check / `current_branch` / `refs/remotes/origin/{branch}` / head / `find_annotated_commit` / marker / `begin_merge` sequence is now in two places, so when the production order changes (e.g. marker written after `merge_analysis`) the hook silently tests a different sequence than ships. Either extract a shared `begin_owned_merge(repo, local_oid, remote_oid, annotated)` that both call, or drop the hook and let the test drive `repo.merge` and write the marker file itself, as the sibling `foreign_merge_without_reflect_marker_is_not_touched` already does.

Either way it needs a `///` comment saying *why* it stops between the marker write and `complete_merge` (AGENTS.md: "Always write documentation for public APIs"; the `#[cfg(test)]` helpers in `db/migrations.rs` set the precedent).

apps/desktop/src-tauri/src/git/merge.rs
```diff
@@ -142,24 +160,23 @@ pub(super) fn merge_remote(root: &Path) -> AppResult<MergeOutcome> {
...
-        .their_label(THEIR_LABEL);
-    repo.merge(&[&annotated], Some(&mut merge_opts), Some(&mut checkout))?;
+    let local_oid = repo.head()?.peel_to_commit()?.id();
+    write_recovery_marker(&repo, local_oid, remote_oid)?;
+    if let Err(error) = begin_merge(&repo, &annotated) {
+        let _ = rollback_owned_merge(&repo, local_oid);
+        return Err(error);
+    }
 
     // From here the repo carries MERGE_* state; a failure that leaves it
```
maccman, line 170: This block now describes the old contract. It sits above a hard reset, not a state clear, and "the next cycle re-derives anything a failed attempt lost" is no longer true for uncommitted edits (see the comment on line 323). `complete_merge`'s doc likewise still says the split exists "so `merge_remote` can guarantee `cleanup_state` runs". Worth rewriting to say the failure path rolls back to the local parent and drops the marker, and what that discards — otherwise the next person extends the pattern assuming it is non-destructive.

maccman:
Thanks for digging into this — the marker-before-merge idea is the right shape, and the iPhone repro is a real problem we want fixed.

Closing this one as-is, though, because in its current form it doesn't reach the failure it describes and adds a few new ways to get stuck (details in the review above). The short version:

- The sync engine runs the commit step before the merge step, and `commit_all` has its own clean-state guard and index write. Both fail on the Merge state / stale `index.lock` this PR targets, so `recover_interrupted_merge` is never called from the real loop.
- The hard reset in `rollback_owned_merge` discards notes edited during the fetch window or after a crash.
- A stale, corrupt, or un-deletable marker becomes a permanent per-cycle error with no self-heal.

A revised version would be welcome. What it needs:

1. Run the recovery where the guard lives (`ensure_clean_state` or a shared repo-open step) so `commit_all` self-heals too, with a test that drives `commit_all → fetch → merge_remote` in engine order.
2. Treat a marker whose commits are already behind HEAD as stale: delete it and continue rather than error.
3. Check `MERGE_HEAD == remote_oid` before rolling back, and make the post-commit marker delete best-effort.
4. Don't hard-reset the whole tree — scope the checkout to paths the merge touched, or preserve files that differ from both sides first.

Stale `index.lock` recovery is a more general problem (it bites `commit_all` and fast-forward pulls with no marker at all) and is better handled separately at repo open.

Happy to review a follow-up.

LemieuxStrategicSolutions:
Thanks for the detailed review. I rebuilt the change around the production engine order and opened the corrected follow-up in #1209. It recovers before commit_all, verifies HEAD/MERGE_HEAD/ORIG_HEAD, resumes in place without a hard reset, and preserves post-crash edits.

PR #1207, opened 2026-09-02
fix: stop iCloud graph watches from pinning fileproviderd

### LINKED ISSUE #1180: iCloud graph causes fileproviderd to sustain ~200% CPU while Reflect is open
## Summary

Opening an iCloud-backed graph in Reflect 0.10.0 causes macOS `fileproviderd` to remain around 185–205% CPU, with Reflect itself around 40–48% CPU. Quitting Reflect makes `fileproviderd` return to 0%; reopening Reflect immediately reproduces the load. Restarting `fileproviderd` only helps until Reflect resumes.

This occurs with a relatively small graph: 95 MB, 54 Markdown files, and 342 files total.

## Environment

- Reflect 0.10.0 stable (`app.reflect.desktop`)
- macOS 15.6.1 (24G90)
- MacBook Pro Mac16,1, Apple M4, 24 GB RAM
- Graph location: Reflect’s iCloud container (`~/Library/Mobile Documents/iCloud~app~reflect/Documents/<graph>`)

## Steps to reproduce

1. Open a graph stored in Reflect’s iCloud container.
2. Leave Reflect open, even without actively editing.
3. Inspect CPU usage in Activity Monitor or with `ps`.
4. Quit Reflect and observe `fileproviderd` drop back to idle.
5. Relaunch Reflect and observe the CPU load return.

## Expected behavior

Reflect and iCloud monitoring should become idle when no files are changing.

## Actual behavior

A representative sample after the behavior had remained active for roughly 30 minutes:

```text
PID    %CPU   COMMAND
19048  193.3  fileproviderd
25649   39.9  reflect-open
```

The result is repeatable: `fileproviderd` stays around 185–205% and Reflect around 40–48%. Running `killall fileproviderd` restarts the daemon, but the high load returns while Reflect is open.

Isolation test:

- Reflect open → sustained high CPU
- Reflect quit → `fileproviderd` returns to approximately 0%
- Reflect reopened → sustained high CPU returns
- The only registered File Provider domains observed were iCloud Drive and Photos

## Diagnostics

A process sample of `fileproviderd` showed repeated filesystem-tree and metadata/xattr work, including `com.apple.fssync.fstree`. A Reflect sample showed File Provider XPC reply handling and file-attribute reads. Reflect also had `.reflect/index.sqlite-shm` open.

The graph’s `.reflect` directory is 4.9 MB and has these relevant xattrs:

```text
com.apple.fileprovider.ignore#P: 1
com.apple.metadata:com_apple_backup_excludeItem
com.dropbox.ignored: 1
```

## Possible trigger

This may be the known excluded-directory interaction described in [`docs/icloud-sync.md`](https://github.com/team-reflect/reflect-open/blob/v0.10.0/docs/icloud-sync.md): the graph-wide `NSMetadataQuery` encounters `.reflect` and `.git`, while those directories are intentionally marked local-only and therefore have no File Provider identity. The documentation notes repeated watch attempts after filesystem events; on this machine, that path appears to produce sustained CPU usage rather than only benign log noise.

Relevant implementation:

- [`icloud/watch.rs`](https://github.com/team-reflect/reflect-open/blob/v0.10.0/apps/desktop/src-tauri/src/icloud/watch.rs)
- [`fs/io.rs`](https://github.com/team-reflect/reflect-open/blob/v0.10.0/apps/desktop/src-tauri/src/fs/io.rs)

Would it be possible to keep `.reflect` outside the iCloud container, or prevent the metadata query from traversing `.reflect` and `.git`? I can provide a focused `sample` or unified-log capture if useful.

maccman:
## Problem

Opening an iCloud-backed graph installs an `NSMetadataQuery` whose path predicate previously covered every descendant of the graph root. Reflect separately marks `.reflect` and `.git` local-only with `com.apple.fileprovider.ignore#P`, so those directories exist on disk but have no File Provider identity.

Filtering delivered metadata items in Rust did not protect File Provider itself: CloudDocs had already attempted to gather and watch the excluded items, and local index or Git writes could make it retry. On affected Macs this leaves `fileproviderd` and Reflect consuming sustained CPU while the graph is idle.

Closes #1180.

## Before -> After

| | Before | After |
|---|---|---|
| Metadata query | Includes every path below the graph root | Excludes `.reflect`, `.git`, and all descendants before File Provider gathers |
| Local-only runtime writes | Can retrigger CloudDocs work for unresolvable items | Stay outside the metadata query while normal note/conflict updates remain covered |
| Path boundaries | Graph root only | Graph root plus exact excluded-directory boundaries for every canonical root spelling |

## Changes

1. Extract `query_predicate` as the single construction point for the live query's provider-side path filter.
2. Exclude both the local-only directory entries and their descendants for the spelled and canonical graph roots, without excluding similarly named paths such as `.reflect-old`.
3. Add a macOS Foundation-backed regression test that evaluates the actual `NSPredicate` across included notes, both excluded trees, canonical root variants, and sibling-boundary cases.
4. Update the iCloud implementation notes to document the corrected metadata-query boundary.

## Tests

- The new predicate test verifies ordinary notes still match under either root spelling.
- It verifies `.reflect`, `.git`, and nested contents do not match.
- It verifies `.reflect-old` remains in scope and a sibling graph with a shared string prefix remains out of scope.
- The complete `icloud::watch` test selection passes (15 tests).

## Verification

- `pnpm check`
- `cargo fmt --check`
- `cargo test -p reflect-open --lib icloud::watch -- --nocapture`
- `pnpm release:macos --flavor=beta --no-notarize` (optimized signed beta build; app signature, embedded-profile identity entitlements, and sidecars verified)
- Launched the signed beta bundle against the existing `reflect-notes` iCloud graph on macOS, confirmed the real `.reflect` and `.git` directories carry `com.apple.fileprovider.ignore#P`, exercised 150 rapid writes under `.reflect`, and observed both the app and `fileproviderd` return to 0% CPU rather than remain pinned.

The reporter's sustained 30-minute runaway did not reproduce reliably with the older installed beta on this Mac, so the live check validates idle recovery and the exact query boundary rather than claiming a full hardware-specific reproduction.

## Risk / Rollout

No migration or data rewrite. The query continues to cover every graph path it covered before except the two trees Reflect already treats as unsynced, hidden runtime state; downstream note classification already rejected both trees. The ordinary macOS `notify` watcher remains unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * iCloud metadata queries now exclude `.reflect` and `.git` directories and their contents, while preserving similarly named paths such as `.reflect-old`.
  * Standard filesystem watching remains active, and iCloud note-state and conflict updates continue to be delivered on Apple platforms.

* **Documentation**
  * Updated iCloud synchronization documentation to describe the metadata-query boundaries and behavior.

### DIFF
--- a/apps/desktop/src-tauri/src/icloud/watch.rs
+++ b/apps/desktop/src-tauri/src/icloud/watch.rs
@@ -159,6 +159,13 @@ mod platform {
     /// of batches.
     const UPDATE_BATCHING_INTERVAL_S: f64 = 2.0;
 
+    /// Graph-local runtime trees that are deliberately excluded from every
+    /// file-sync provider. They have no File Provider identity, so asking a
+    /// ubiquitous metadata query to watch them makes CloudDocs repeatedly
+    /// retry an item it can never resolve. Keep the provider-side query off
+    /// these paths; filtering results later is too late to avoid that work.
+    const SYNC_EXCLUDED_DIRS: [&str; 2] = [".reflect", ".git"];
+
     /// The live query plus everything that must stay alive (and on the main
     /// thread) with it.
     struct Watch {
@@ -315,6 +322,49 @@ mod platform {
         variants
     }
 
+    /// Build the provider-side path predicate for one graph. Each root is
+    /// slash-terminated by [`root_variants`], so the positive clauses cannot
+    /// claim a sibling graph. The exclusions cover both the directory item
+    /// itself and its descendants without blacking out similarly named user
+    /// paths such as `.reflect-old`.
+    fn query_predicate(roots: &[String]) -> Retained<NSPredicate> {
+        let path_key: Retained<NSString> = unsafe { NSMetadataItemPathKey.copy() };
+        let included = (0..roots.len())
+            .map(|_| "(%K BEGINSWITH %@)")
+            .collect::<Vec<_>>()
+            .join(" OR ");
+        let exclusions = roots
+            .iter()
+            .flat_map(|_| SYNC_EXCLUDED_DIRS)
+            .map(|_| "(%K != %@) AND (NOT (%K BEGINSWITH %@))")
+            .collect::<Vec<_>>()
+            .join(" AND ");
+        let format = NSString::from_str(&format!("({included}) AND {exclusions}"));
+
+        let mut arg_list: Vec<Retained<NSString>> = Vec::new();
+        for root in roots {
+            arg_list.push(path_key.copy());
+            arg_list.push(NSString::from_str(root));
+        }
+        for root in roots {
+            for directory in SYNC_EXCLUDED_DIRS {
+                let excluded = format!("{root}{directory}");
+                arg_list.push(path_key.copy());
+                arg_list.push(NSString::from_str(&excluded));
+                arg_list.push(path_key.copy());
+                arg_list.push(NSString::from_str(&format!("{excluded}/")));
+            }
+        }
+        let args = NSArray::from_retained_slice(&arg_list);
+        unsafe {
+            msg_send![
+                objc2::class!(NSPredicate),
+                predicateWithFormat: &*format,
+                argumentArray: &*args
+            ]
+        }
+    }
+
     /// Build, wire, and start the query. Main thread only. Tears down any
     /// live watch first (installs and stops all run here, serially), and
     /// aborts when a later `start`/`stop` has superseded this one's epoch —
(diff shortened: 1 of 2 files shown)


maccman:
Closing in favor of #1208. Its entitled CloudDocs probe showed that predicate exclusions do not constrain the container-wide gather, so #1208's desktop-only removal of NSMetadataQuery addresses the root cause more reliably.

PR #1208, opened 2026-09-02
fix: stop the iCloud metadata query from pinning fileproviderd on macOS

### LINKED ISSUE #1180: iCloud graph causes fileproviderd to sustain ~200% CPU while Reflect is open

maccman:
Fixes #1180.

## Problem

Opening an iCloud-backed graph on macOS 15.6 held `fileproviderd` around 185–205% CPU and Reflect around 40% for as long as the app stayed open, with a 54-note graph and no edits. Quitting Reflect dropped `fileproviderd` back to idle; relaunching brought the load straight back.

## Root cause

The desktop app installed the same `NSMetadataQuery` watch mobile uses, as a conflict signal. That query's cost is not confined to the graph. Measured on macOS 26.5 with a small entitled probe app over the real container (the exact shape `icloud/watch.rs` builds — ubiquitous documents scope, path-prefix predicate, serial delivery queue), with CloudDocs debug logging on:

- CloudDocs' `BRItemCollectionGatherer` enumerates the app's **entire** ubiquity container and builds one observed collection per directory, whatever the predicate says.
- Every sync-excluded directory in the container — `.reflect/` and `.git/` in *every* graph it holds — carries no provider identity, so its collection fails with `NSFileProviderInternalErrorDomain 15 while gathering`, is retried with a 10→50 ms backoff, and is given up on after six rounds (`[CRIT] UNREACHABLE: … BRItemCollectionGatherer - Repeatedly can't watch item`). Every retry reloads the whole item tree through `fileproviderd`: a three-graph container of ~13k items was re-gathered six times per query start. Churn inside an excluded tree re-triggers the failure (a freshly created one produced 35 give-ups in a single run).
- Nothing scopes the gather below the container. `NOT (path BEGINSWITH …/.reflect/)` predicates: identical failure counts. `searchItems`: ignored. URL search scopes: routed to Spotlight, zero results, no ubiquitous attributes. Marking the directory via the `NSURLUbiquitousItemIsExcludedFromSyncKey` resource key (which sets the same `com.apple.fileprovider.ignore#P` xattr) or naming it `.nosync`: still one failing collection each.

macOS 26 bounds the retry cycle at those six rounds and the query then goes silent; the reporter's sample (`com.apple.fssync.fstree` walks and xattr reads in `fileproviderd`, File Provider XPC replies in Reflect) is that same reload cycle never settling on macOS 15.6. The "benign log noise" note in `docs/icloud-sync.md` was describing this fault from the wrong side.

## Fix

Desktop never needed the query for freshness — the `notify` watcher already reports every iCloud write as a plain FS event — only for the conflict signal and sweep scoping. So the watch becomes mobile-only, where it is the sole external-change source and iOS has no file watcher to fall back on.

- `createIcloudController` takes `metadataWatch` (true on mobile). Desktop no longer installs, restarts, or subscribes to the watch; it mounts the sweeps alone.
- Desktop arrival sweeps get a new `'ingested'` scope (`SweepScope::Ingested`): the `NSFileVersion` checks cover the notes that just arrived, instead of degrading to the full per-note pass that the `'candidates'` scope falls back to without a watch. A remote edit that became the working file is by definition among the arrivals. A conflict version that never touches the working file waits for the resume/focus full sweep — the same backstop graphs kept outside the app's container always relied on.
- `icloud_watch_start` drops its `emit_file_changes` flag; the watch always emits now.
- `docs/icloud-sync.md` records the measured mechanism.

## Testing

- `pnpm check` (typecheck, oxfmt, oxlint, eslint) clean; `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `pnpm test --run apps/desktop/src/lib/icloud-controller.test.tsx`: 19 pass, including new regression tests — desktop never invokes `icloud_watch_start`/`icloud_watch_stop` (start, resume, dispose) and never subscribes to the watch's events; desktop arrival sweeps carry `scope: 'ingested'` with the arrived paths; a failed desktop arrival sweep retries at full scope; the mobile shape keeps its watch lifecycle and candidate scoping.
- `cargo test -p reflect-open --lib -- icloud`: 42 pass, including the new `conflict_candidates` test (`Ingested` → exactly the arrivals, `Full` → everything, `Candidates` without a watch → degrades to everything, never to an empty set).
- Probe measurements above (macOS 26.5.1, Apple Silicon, the real `iCloud.app.reflect` container with three graphs).
- App-level run of a signed build of this branch against a graph in the real container: zero `NSMetadataQuery`/gatherer activity from Reflect's process, `fileproviderd` at 0.0% throughout, baseline sweep at `Full` and both arrival sweeps at `Ingested` — full numbers in the first comment below.

## Not covered

- The mechanism itself could only be measured on macOS 26; a macOS 15.6 machine is needed to confirm the reload cycle now ends there, which the fix guarantees by construction (no query, no gatherer in Reflect's process).
- iOS keeps the watch, so an iPhone still pays the bounded gather cost per watch start; if an iOS-side CPU report ever comes in, the answer is structural (`.reflect` outside the container, or an `NSFilePresenter`-based conflict signal), not a predicate.
- Desktop promptness for a conflict version that never touches the working file now depends on the next refocus rather than the query's update round.

### DIFF
--- a/apps/desktop/src-tauri/src/icloud/sweep.rs
+++ b/apps/desktop/src-tauri/src/icloud/sweep.rs
@@ -79,9 +79,28 @@ pub enum SweepScope {
     /// ([`crate::icloud::watch::conflicted_paths`]) — each check is a
     /// synchronous file-coordination round-trip to `fileproviderd`, and a
     /// bulk-sync arrival sweep over a large graph paid one per note for a
-    /// listing the query already has. Degrades to `Full` whenever the watch
-    /// cannot answer completely (not installed, or pre-gather).
+    /// listing the query already has. Mobile's signal and arrival sweeps.
+    /// Degrades to `Full` whenever the watch cannot answer completely (not
+    /// installed, or pre-gather).
     Candidates,
+    /// Restrict the version checks to this sweep's `ingested_paths` — the
+    /// notes external writes just landed on. Desktop's arrival sweeps: no
+    /// metadata watch runs there (see [`crate::icloud::watch`]), and a
+    /// conflict whose remote side became the working file is by definition
+    /// among the arrivals. A conflict version that never touches the working
+    /// file is the resume sweep's (`Full`) job, exactly as for graphs kept
+    /// outside the app's container.
+    Ingested,
+}
+
+/// The version-check candidate set for `scope` ([`run_sweep`]'s
+/// `conflict_candidates`): `None` checks every note.
+fn conflict_candidates(scope: SweepScope, ingested_paths: &[String]) -> Option<HashSet<String>> {
+    match scope {
+        SweepScope::Full => None,
+        SweepScope::Candidates => crate::icloud::watch::conflicted_paths(),
+        SweepScope::Ingested => Some(ingested_paths.iter().cloned().collect()),
+    }
 }
 
 /// Command: run a conflict sweep over the generation-pinned graph.
@@ -105,10 +124,7 @@ pub async fn icloud_conflicts_scan(
     let root = crate::fs::root_for_generation(&state, generation)?;
     let sweep_root = root.clone();
     let outcome = crate::blocking::run_blocking(move || {
-        let candidates = match scope {
-            SweepScope::Full => None,
-            SweepScope::Candidates => crate::icloud::watch::conflicted_paths(),
-        };
+        let candidates = conflict_candidates(scope, &ingested_paths);
         run_sweep(
             &sweep_root,
             &skip_paths,
@@ -964,6 +980,26 @@ mod tests {
         // …and the next full sweep prunes it.
         run_sweep(root.path(), &[], &[], false, None).unwrap();
         assert!(ShadowStore::new(root.path()).base("notes/a.md").is_none());
+    }
+
+    #[test]
+    fn ingested_scope_checks_exactly_the_arrivals_and_full_checks_everything() {
+        let arrivals = vec!["notes/a.md".to_string(), "daily/2026-07-04.md".to_string()];
+        assert_eq!(
+            super::conflict_candidates(SweepScope::Ingested, &arrivals),
+            Some(arrivals.iter().cloned().collect::<HashSet<_>>())
+        );
+        assert_eq!(
+            super::conflict_candidates(SweepScope::Full, &arrivals),
+            None
+        );
+        // No metadata watch is installed in a test process (nor ever on
+        // desktop): the candidate scope must degrade to a full check, never
+        // to an empty one.
+        assert_eq!(
+            super::conflict_candidates(SweepScope::Candidates, &arrivals),
+            None
+        );
     }
 
     #[test]
--- a/apps/desktop/src-tauri/src/icloud/watch.rs
+++ b/apps/desktop/src-tauri/src/icloud/watch.rs
@@ -1,15 +1,29 @@
 //! The iCloud change watcher: an `NSMetadataQuery` over the graph (Plan 21
-//! Phase 2).
+//! Phase 2). **Mobile only.**
 //!
-//! Two jobs, per platform:
+//! On iOS this query is the *sole* external-change source — there is no file
+//! watcher on mobile — so its snapshot diffs become the standard
+//! `index:changed` batches the indexer and open sessions already consume,
+//! and its `HasUnresolvedConflicts` flags are the conflict signal that
+//! triggers a prompt, candidate-scoped sweep.
 //!
-//! - **iOS**: the *sole* external-change source. There is no file watcher on
-//!   mobile — this query's snapshot diffs become the standard `index:changed`
-//!   batches the indexer and open sessions already consume.
-//! - **Both Apple platforms**: the conflict signal. A conflict version
-//!   appearing does not necessarily touch the working file, so the desktop
-//!   `notify` watcher alone would sit silent; the query's
-//!   `HasUnresolvedConflicts` flag is what triggers a sweep promptly.
+//! Desktop deliberately never installs it (issue #1180). The `notify` watcher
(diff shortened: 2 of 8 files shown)


maccman:
## App-level verification (macOS 26.5.1, real `iCloud.app.reflect` container)

Built this branch as the stable flavor (`pnpm tauri build --target aarch64-apple-darwin --bundles app`), signed it with the Developer ID certificate plus the stable provisioning profile's identity entitlements (so the iCloud entitlement is live), and launched it beside the running Reflect Beta with a private `HOME` whose recents pointed at the **Demo Graph** in the container. CloudDocs/FileProvider debug logging was captured for the process (`log stream --process reflect-open --predicate 'processIdentifier == <pid>' --info --debug`), and an external note was written into `notes/` and removed again while it ran.

**Process log (2,392 lines over ~2 minutes):**

| Signature | Count |
| --- | ---: |
| `NSMetadataQuery` / `startQuery` / `BRQuery` | 0 |
| `BRItemCollectionGatherer` | 0 |
| `NSFileProviderInternalErrorDomain … while gathering` | 0 |
| `UNREACHABLE … Repeatedly can't watch item` | 0 |
| `_buildCollectionOnItemIfPossible` (item-tree rebuilds) | 0 |
| `FPDaemonConnection valuesForAttributes:forItemAtURL:` (placeholder attribute reads) | 10 |

`fileproviderd` sat at **0.0% CPU** before, during, and after the external write; so did `reflect-open` at idle.

**App tracing (stdout):**

```text
18:44:41 watcher: watch_start installed
18:44:42 icloud_conflicts_scan … record_baseline=true scope=Full elapsed_ms=31      ← open
18:45:43 db: index_reconcile_scan files=9 candidates=1                              ← external note landed
18:45:46 icloud_conflicts_scan … record_baseline=false scope=Ingested elapsed_ms=24 ← arrival sweep, arrivals only
18:46:31 db: index_reconcile_scan files=8                                           ← external note removed
18:46:32 icloud_conflicts_scan … record_baseline=false scope=Ingested elapsed_ms=1
```

For contrast, the same container measured with an entitled `NSMetadataQuery` probe (the shape this branch removes from desktop) produced 36 `while gathering` errors, 30 collection restarts, 6 give-ups, and ~87k item-tree rebuild lines per query start — see the PR description.

The probe note and its removal were also committed by the Demo Graph's local Git history engine, as any external edit would be; nothing else was left behind.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

ocavue (approved):

PR #1209, opened 2026-09-02
fix: sync no longer gets stuck after the app is killed mid-merge

LemieuxStrategicSolutions:
## Problem

Follow-up to #1205.

If the process is terminated after libgit2 enters merge state, the next sync currently reaches `commit_all` first and fails its clean-state guard. The device remains at Needs attention. A stale index lock from the interrupted owned merge can block the same path.

## Solution

- atomically record the two merge parents before entering libgit2 merge state
- run recovery in a shared repository-open preflight used by status, commit, fetch, merge, and push
- resume only when HEAD, MERGE_HEAD, and ORIG_HEAD exactly match the app-owned marker
- continue the existing merge index in place; never hard-reset the working tree
- preserve post-crash edits so the immediately following commit captures them
- keep status informative when recovery fails while write paths remain blocked
- treat clean, corrupt, unsupported, or superseded markers as stale metadata
- make marker deletion best-effort after durable completion
- leave foreign CLI merges/rebases and their locks untouched

A generic stale `index.lock` without an app-owned marker remains out of scope and should be handled separately.

## Review feedback addressed

- recovery now runs before `commit_all`
- the end-to-end test follows `commit_all → fetch → merge_remote → push`
- no whole-tree reset
- MERGE_HEAD and ORIG_HEAD are verified
- stale or invalid markers self-heal
- completed-merge cleanup cannot clear a newer foreign operation
- marker writes reuse `atomic_write_to`
- interrupted cleanup after the durable merge commit has direct coverage

## Validation

- `pnpm check`
- `cargo fmt --check`
- `cargo clippy -p reflect-open --all-targets -- -D warnings`
- `cargo test -p reflect-open` — 377 passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Interrupted merges now resume automatically when the repository is reopened.
  - Edits made after an interruption are preserved during conflict resolution.
  - Recovery no longer removes an active index lock, preventing interference with other Git operations.
  - Commit, fetch, pull, and push operations now complete pending merge recovery before proceeding.
  - Repository status remains available even when interrupted-merge recovery cannot complete.

### DIFF
--- a/apps/desktop/src-tauri/src/git/commit.rs
+++ b/apps/desktop/src-tauri/src/git/commit.rs
@@ -9,7 +9,7 @@ use serde::Serialize;
 use crate::error::AppResult;
 
 use super::commit_message::message_for_commit;
-use super::repo::{ensure_clean_state, open_existing, signature};
+use super::repo::{ensure_clean_state, open_for_sync, signature};
 
 /// A file whose *changes* were withheld from staging because it is at/above
 /// the size guardrail. Oversized-but-unchanged files are not reported — their
@@ -45,7 +45,7 @@ pub(super) fn commit_all(
     fallback_message: &str,
     max_file_bytes: u64,
 ) -> AppResult<CommitOutcome> {
-    let repo = open_existing(root)?;
+    let repo = open_for_sync(root)?;
     ensure_clean_state(&repo)?;
     // In-memory hard guarantee, independent of any on-disk ignore file: the
     // runtime directory must never enter a backup commit (Plan 21 — a synced
--- a/apps/desktop/src-tauri/src/git/merge.rs
+++ b/apps/desktop/src-tauri/src/git/merge.rs
@@ -246,7 +246,7 @@ pub(super) fn recover_interrupted_merge(repo: &Repository) -> AppResult<()> {
     if merge_commit_has_parents(repo, head_oid, local_oid, remote_oid)? {
         if repo.state() == git2::RepositoryState::Merge && refs_match {
             // The merge commit landed and only cleanup was interrupted.
-            remove_owned_index_lock(repo)?;
+            ensure_index_unlocked(repo)?;
             repo.cleanup_state()?;
         }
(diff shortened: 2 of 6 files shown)


LemieuxStrategicSolutions:
Follow-up commit `81f21649` addresses the two remaining data-integrity concerns:

- Recovery no longer deletes `.git/index.lock` based on merge-marker ownership. An unknown lock is left intact and recovery waits, preventing a race with another Git process.
- Edit-vs-delete and binary conflict recovery no longer overwrite an existing working-tree file, so edits made after interruption are staged into the recovered merge.
- Added regression coverage for a foreign lock, post-crash edits, and the durable-commit / interrupted-cleanup path.

Local validation is clean:

- `pnpm check`
- `cargo fmt --all --check`
- `cargo clippy -p reflect-open --all-targets -- -D warnings`
- `cargo test -p reflect-open` (379 passed)

@maccman, this is the focused follow-up to your feedback on #1205. When you have a chance, would you please review the narrowed recovery fix?

LemieuxStrategicSolutions:
Closing this approach in favor of #1215's in-memory merge. I compared its current head (`47804816`) with this PR: `merge_commits` computes the result without entering libgit2 merge state, and checkout/index writes complete before the branch ref advances. That avoids the new app-owned `MERGE_*` state this recovery mechanism was designed to repair.

The final review on this PR still identified post-interruption deletion and binary conflict-copy preservation cases; I am not marking those as fixed in this implementation. Instead, #1218 contributes two focused failure/retry tests directly against #1215, checking that an edited recovery copy survives and that a later deletion is snapshotted before a fresh edit/delete merge. The latter retains the existing conflict policy and verifies the deletion remains in the local merge parent.

All 11 failure/retry tests passed locally, along with formatting, strict clippy, `pnpm check`, and `pnpm build`. #1215 and #1218 are still awaiting upstream acceptance; this closure does not mean either has shipped. Existing unfinished foreign Git operations remain outside automatic recovery scope.

PR #1215, opened 2026-09-04
fix: preserve edits and recovery data during Git sync

maccman:
Fixes a data-loss window where a note saved during fetch could be overwritten by a forced fast-forward checkout. Reflect now snapshots those saves under the same local mutation lock used by checkout. Network waits leave editing responsive; saves queued during checkout land afterwards, and delayed editor reads cannot replace the newly saved buffer.

The same pull path now computes merges in memory, uses safe checkout before moving branch refs, preserves unrelated staging, and retains recoverable history and complete conflict copies when applying a pull fails. Binary conflicts claim numbered filenames atomically, including protection against existing files and incoming case aliases.

Runtime exclusion now removes previously tracked `.reflect` entries (including case and trailing-dot/space aliases) from future commits and filters incoming checkouts and restores without deleting local SQLite or durable chat data. Pre-staged oversized files cannot bypass the attachment size guard. Backup lifecycle fixes preserve local commits on auth failure, push saves captured after fetch, await queued manual cycles, and prevent stale startup/disconnect work from reviving an old controller. Native deletion events with null timestamps now reach the editor and complete the sync cycle. Imports and renames share the checkout boundary, including cancellation and native event-loop handling.

Validation:

- Reproduced all three original audit failures against current master before editing.
- `pnpm check` and `pnpm build` passed.
- 161 targeted TypeScript tests passed across sync command boundaries, engine, backup controller, note session, and document binding.
- `cargo fmt --all --check` and `cargo clippy -p reflect-open --all-targets -- -D warnings` passed.
- 288 targeted Rust tests passed: Git (60), filesystem (105), index (64), iCloud (41), and capture (18). Desktop sidecars were staged first.
- Regression coverage includes saves during fetch/checkout, dirty and ignored checkout collisions, failed index writes, repeated/concurrent binary conflicts, tracked runtime adoption/pull/clone and Windows filename aliases, native null deletion timestamps, and stale editor reads after a save.

Existing Git history is not rewritten, so runtime blobs already committed in old revisions remain there. The mutation lock coordinates Reflect's own writers; external filesystem tools do not participate. An I/O failure may leave some pulled files applied or a complete conflict copy on disk while the previous branch tip remains recoverable. These boundaries are documented in `docs/git-backup-safety.md`.

Supersedes #1214, which was produced by the previously cancelled duplicate CLI run. Its branch is preserved. Bugbot is disabled for this repository; the repository’s CodeRabbit review remains enabled.

## Summary by CodeRabbit

- **New Features**
  - Improved Git backup and sync behavior, including safer merges, conflict-copy preservation, and clearer reporting when oversized files are skipped.
  - Local runtime data remains protected during cloning, merging, and remote operations.
  - Local history can continue saving even when remote credentials or connectivity are unavailable.

- **Bug Fixes**
  - Prevented stale external file updates from overwriting newer editor changes.
  - Improved reliability for concurrent saves, imports, note moves, and graph operations.
  - Added safeguards for unresolved Git conflicts and interrupted operations.

- **Documentation**
  - Added guidance for Git backup safety, recovery, conflict handling, and runtime-file protection.

### DIFF
--- a/apps/desktop/src-tauri/src/capture.rs
+++ b/apps/desktop/src-tauri/src/capture.rs
@@ -498,8 +498,11 @@ pub async fn capture_screenshot_promote(
     state: State<'_, GraphState>,
 ) -> AppResult<()> {
     let root = root_for_generation(&state, generation)?;
-    let bytes = fs::read(inbox_file(&root, &spool_name)?)?;
-    let jpeg = downscale_jpeg(&bytes, max_dim)?;
+    let jpeg = crate::blocking::run_blocking(move || {
+        let bytes = fs::read(inbox_file(&root, &spool_name)?)?;
+        downscale_jpeg(&bytes, max_dim)
+    })
+    .await?;
     crate::fs::mutation::run(state.inner().clone(), generation, move |root| {
         persist_asset(root, &asset_path, &jpeg)
     })
--- a/apps/desktop/src-tauri/src/capture.rs
+++ b/apps/desktop/src-tauri/src/capture.rs
@@ -490,17 +490,20 @@ fn persist_asset(root: &Path, asset_path: &str, bytes: &[u8]) -> AppResult<()> {
 /// not move — the drain removes spool files only after the note is written,
 /// so a crash mid-drain re-runs cleanly.
 #[tauri::command]
-pub fn capture_screenshot_promote(
+pub async fn capture_screenshot_promote(
     spool_name: String,
     asset_path: String,
     max_dim: u32,
     generation: u64,
-    state: State<GraphState>,
+    state: State<'_, GraphState>,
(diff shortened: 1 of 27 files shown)


maccman:
Addressed the screenshot-worker finding in 47804816: spool I/O and JPEG decoding now use `blocking::run_blocking`; only final asset persistence holds the graph mutation lock. The generation is still rechecked at that write boundary.

Validation passed: 18 capture tests, desktop Clippy with warnings denied, Rust formatting, `pnpm check`, and `pnpm build`.

The docstring coverage warning includes private helpers and tests. Public contracts and the non-obvious locking/recovery behavior are documented; I have kept the repository convention against comments that only restate code.

PR #1218, opened 2026-09-05
test: preserve user changes across failed pull retries

LemieuxStrategicSolutions:
Adds two failure/retry regressions to #1215, following the remaining review cases from #1209:

- After a failed checkout leaves a binary conflict copy, a user edits that copy. The next snapshot and pull must retain those bytes in the working tree and merge tree, and preserve the remote attachment in a separately named copy.
- After a failed checkout, a user deletes the conflicted note. The next snapshot must commit that deletion before a new edit/delete merge. The test explicitly preserves the existing keep-the-edited-version policy: the remote note may return as a reported conflict, while the deletion remains in the local merge parent.

This is intentionally based on `codex/git-sync-reliability` so the diff contains only the added tests. It can be merged into #1215 or cherry-picked; it does not ask to merge the old recovery mechanism from #1209. The in-memory merge in #1215 avoids creating the `MERGE_*` state that mechanism would recover.

Validation passed: `cargo test -p reflect-open git::merge_tests` (11 tests, including both additions), `cargo fmt --all --check`, `cargo clippy -p reflect-open --all-targets -- -D warnings`, `pnpm check`, and `pnpm build`. These tests exercise deterministic checkout failure and retry; they do not claim physical-device kill testing or migration of pre-existing foreign Git merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merge recovery after a failed checkout when the worktree contains uncommitted changes.
  * Preserved edits to conflict copies and recorded file deletions before retrying a merge.
  * Ensured retried merges retain local changes while correctly surfacing incoming content conflicts.

### DIFF
--- a/apps/desktop/src-tauri/src/git/merge_tests.rs
+++ b/apps/desktop/src-tauri/src/git/merge_tests.rs
@@ -408,3 +408,109 @@ fn binary_copy_avoids_incoming_case_aliases_and_directories() {
         );
     }
 }
+
+#[test]
+fn retry_preserves_a_conflict_copy_edited_after_failed_checkout() {
+    let fixture = fixture();
+    write(&fixture.origin, "notes/shared.md", "remote note\n");
+    write(&fixture.origin, "assets/image.bin", b"remote attachment\0");
+    commit(&fixture.origin);
+    write(&fixture.local, "notes/shared.md", "local note\n");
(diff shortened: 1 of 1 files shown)


LemieuxStrategicSolutions:
@coderabbitai review

PR #1246, opened 2026-09-08
feat(capture): save new X bookmarks with public text

ocavue:
Save new X bookmark actions into Reflect after an optional browser host-permission grant. The extension observes `CreateBookmark` requests and sends ordinary v1 link envelopes through the existing queue and native host.

New X captures create independent UUID notes and repair their Daily backlink on replay without rebuilding saved content. Reflect saves the URL and manual input first, then appends available public text through the existing bounded JSON transport. Private, dirty, edited, or changed captures stop automatic writes. Supplied page text completes the capture directly; X captures do not run AI, link previews, or automatic retitling.

Scope: bookmark intent only, including requests X later rejects. No DOM extraction, likes, historical import, quote expansion, or media download. Long posts may be previews. The existing 50-entry queue limit still applies. Upgrade desktop before opting in; the unchanged wire format does not change old desktop behavior.

Validation:
- Added request/permission, Unicode/text normalization, create/replay, cleanup failure, privacy/edit-race, and terminal-write tests.
- After the initial CI failure, all 154 targeted tests, TypeScript, changed-file lint, and the Chrome MV3 production build passed locally. Full CI is running on the updated head.
- Read-only live syndication probes for three public IDs returned matching IDs within the existing 64 KiB response cap, including an empty display-text range.
- Draft remains open for live signed-in X button/menu/shortcut, worker wake/revocation, and app-closed capture acceptance. Offline fixtures do not establish those live contracts.

Change size relative to master: production files +778/-51 (net +727), tests net +559, documentation net +31, configuration net +3. Production counts include comments, whitespace, and inline Rust tests. No new dependency or wire schema was added.

### DIFF
--- a/apps/desktop/src-tauri/src/capture.rs
+++ b/apps/desktop/src-tauri/src/capture.rs
@@ -534,22 +534,15 @@ pub async fn capture_meta_fetch(url: String) -> AppResult<String> {
     Ok(String::from_utf8_lossy(&response.body).into_owned())
 }
 
-// ---- oEmbed fetch ---------------------------------------------------------------
+// ---- JSON fetch -----------------------------------------------------------------
 
-/// oEmbed answers are ~1 KB of JSON; anything past this cap is not an oEmbed
-/// answer, and a truncated one would be unparseable, so oversize is an error
-/// rather than a cut.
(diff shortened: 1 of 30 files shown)


ocavue:
implemented.

PR #1342, opened 2026-09-21
fix(desktop): open new-window requests in the browser

ocavue:
A `target="_blank"` link or `window.open` call, including one inside an embedded player such as the "Watch video on YouTube" button, did nothing because the webview had no new-window handler. The main window is now built in `setup` so it can carry one, and http(s) requests open in the default browser.


ocavue (commented):
Reviewed by Codex (OpenAI).

No actionable findings in this revision. I checked the main-window startup and recreation paths, secondary-window handling, configuration overlays, saved-window-state restoration, and the popup callback against the pinned Tauri/Wry source.

All reported CI checks pass. This was a source review; I did not launch the native app or manually exercise the embedded YouTube player. The new-window callback is unsupported on iOS/Android in the pinned dependency, so this remains a desktop fix.

